PR #3, opened 2026-08-25
Keep Hyprland toggle catalog out of new-user state

### LINKED ISSUE #1: Tiling window manager
Amazing to get this on my M4 MacBook! Quick question though: why is the first window that you open, opened centered in the screen, not full screen? And there also seem to be no gaps/borders between windows/tiles. Is this a different Hyprland setup than the default Omarchy setup you'd install normally?

Keen to hear back!

pfelrodrigues:
Fixes #1

`omarchy-hyprland-toggle` treats any file in `~/.local/state/omarchy/toggles/hypr/` as an enabled flag. `materialize-omarchy.sh` was copying the catalog from `default/hypr/toggles/` into that directory for every new user.

That turned on:

- `window-no-gaps` (gaps and borders at 0)
- `single-window-aspect-ratio` (first window as a centered 1:1 square)

The catalog stays under `/usr/share/omarchy/default/hypr/toggles/` so users can still toggle the flags. New sessions start with an empty state directory.

Existing VMs already copied those files into `$HOME`. A factory reset (or deleting those two lua files from the toggle state dir) is needed to pick this up.

### DIFF
--- a/guest/scripts/materialize-omarchy.sh
+++ b/guest/scripts/materialize-omarchy.sh
@@ -147,8 +147,9 @@ copy_tree "$source_dir/config" "$root/etc/skel/.config"
 install_file 0644 "$source_dir/default/bashrc" "$root/etc/skel/.bashrc"
 mkdir -p "$root/etc/skel/.local/share/applications"
 copy_contents "$source_dir/applications" "$root/etc/skel/.local/share/applications"
+# Presence of a file here enables that Hyprland flag. default/hypr/toggles is
+# the catalog (window-no-gaps, 1:1 single window), not the factory default.
 mkdir -p "$root/etc/skel/.local/state/omarchy/toggles/hypr"
-copy_contents "$source_dir/default/hypr/toggles" "$root/etc/skel/.local/state/omarchy/toggles/hypr"
 
 if [[ -d $source_dir/default/nautilus-python/extensions ]]; then
   copy_tree "$source_dir/default/nautilus-python/extensions" "$root/etc/skel/.local/share/nautilus-python/extensions"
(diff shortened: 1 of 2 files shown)


guest/scripts/materialize-omarchy.sh
```diff
@@ -147,8 +147,9 @@ copy_tree "$source_dir/config" "$root/etc/skel/.config"
 install_file 0644 "$source_dir/default/bashrc" "$root/etc/skel/.bashrc"
 mkdir -p "$root/etc/skel/.local/share/applications"
 copy_contents "$source_dir/applications" "$root/etc/skel/.local/share/applications"
+# Presence of a file here enables that Hyprland flag. default/hypr/toggles is
+# the catalog (window-no-gaps, 1:1 single window), not the factory default.
 mkdir -p "$root/etc/skel/.local/state/omarchy/toggles/hypr"
-copy_contents "$source_dir/default/hypr/toggles" "$root/etc/skel/.local/state/omarchy/toggles/hypr"
```
themartiano, line 151: Here we stop copying the whole directory, but `flags.lua` is also there and not the cause of the issue.
Not sure what's the impact, but to keep Omarchy as close to source as possible, can you please import it here?

themartiano (commented):
A lot of people noticed this, thanks for fixing!

guest/scripts/materialize-omarchy.sh
pfelrodrigues: Done. New sessions now get only `flags.lua` (the empty-directory sentinel). `window-no-gaps` and `single-window-aspect-ratio` stay in the catalog.

mwdossantos: Amazing work guys!

PR #8, opened 2026-08-26
Share the Mac clipboard with Omarchy

### LINKED ISSUE #10: Clipboard sharing between macOS and Omarchy
Text and images copied on the Mac are not available inside Omarchy, and vice versa. Every exchange currently goes through a file or retyping.

**Expected:** copying text or a screenshot on either side makes it available on the other, without focus tricks and without echoing updates back and forth.

**Notes:** the Cocoa display backend has no clipboard bridge for this setup, and SPICE vdagent does not work reliably with Hyprland's Wayland clipboard. The existing native audio bridge (virtserialport + JSON lines + guest agent) is a good pattern to follow; `wl-clipboard` is already in the guest image and its data-control support works on Hyprland without focus.

Implemented in #8.

rlimberger:
Closes #10

## Summary

- Two-way clipboard sharing for text and PNG images between macOS and the Omarchy desktop.
- Host side follows the existing native audio-bridge pattern: a new virtserialport `dev.tryomarchy.clipboard` backed by a private socket, served by `omarchy-vm-helper --bridge-native-clipboard`, which polls `NSPasteboard.changeCount`. The bridge is supervised non-fatally (restarted up to 5×) so a clipboard hiccup never stops the VM. The socket connect/write code shared with the audio bridge moves to `NativeBridgeSocket.swift`.
- Guest side: `omarchy-native-clipboard-bridge` (Python, uses the already-shipped `wl-clipboard`, whose data-control support works on Hyprland without focus) runs as a user unit bound to `graphical-session.target`; a udev rule grants the `users` group the port.
- Both sides fingerprint what they last wrote so nothing echoes back; the guest asks for the current Mac clipboard on login. Payloads are capped at 16 MiB. Unknown formats are ignored rather than fatal so the protocol can grow.
- Runtime contract (`spec.json`), launcher validator, and `guest/tests/verify.py` cover the port, unit, rule, and message schema.

No new host or guest dependencies; no QEMU changes.

## Test plan

- [x] `swift test` (48 tests) and `guest/test` pass
- [x] Copy text on the Mac → paste in Omarchy; copy in Omarchy → `pbpaste` on the Mac
- [ ] Screenshot copied on the Mac pastes as PNG in the guest (implemented and unit-tested; not yet exercised by hand)
- [x] Verified on a full `make build` + first-boot provisioning on Apple Silicon

Note: building on current Homebrew (sdl2-compat/SDL3) needs #6 first; this PR does not touch those pins.

https://claude.ai/code/session_01GopAGVNHBeeDUwEQ21Qotf

### DIFF
--- a/guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
+++ b/guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
@@ -89,10 +89,13 @@ def decode_message(line: bytes) -> tuple[str, bytes] | None:
 class ClipboardSync:
     """Echo-safe two-way state shared by the tests and the live agent.
 
-    Each direction remembers the fingerprint of its last write for a short
-    window. Without the expiry a marker would outlive the content it
-    protects: guest copies B, Mac copies A, Mac copies B again, and the final
-    B would be mistaken for the original echo and lost.
+    A marker only protects content that is still on the other side's
+    clipboard. Accepting a new change in one direction therefore clears the
+    other direction's marker: after guest copies B and Mac copies A, the Mac
+    pasteboard no longer holds B, so a later Mac copy of B is new content and
+    must not be mistaken for the original echo. Echoes are delivered in order
+    before any later user copy, so clearing cannot let one through. A short
+    expiry backs this up for markers that never see an opposite change.
     """
 
     def __init__(self, send: Any, copy: Any, clock: Any = time.monotonic) -> None:
@@ -117,6 +120,7 @@ class ClipboardSync:
         if self._is_echo(self.last_from_host, key, now):
             return False
         self.last_from_guest = (key, now)
+        self.last_from_host = None
         self.send(encode_message(mime, payload))
         return True
 
@@ -127,6 +131,7 @@ class ClipboardSync:
         if self._is_echo(self.last_from_guest, key, now):
             return False
         self.last_from_host = (key, now)
+        self.last_from_guest = None
         self.copy(mime, payload)
         return True
 
--- a/guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
+++ b/guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
@@ -1,0 +1,287 @@
+#!/usr/bin/python3
+"""Share the Wayland clipboard with the macOS pasteboard over a virtio port.
+
+Both directions travel as one JSON object per line on the
+dev.tryomarchy.clipboard virtserialport:
+
+    {"type": "clipboard", "format": "<mime>", "data": "<base64>"}
+
+The host answers {"type": "sync"} with its current pasteboard. Text and PNG
+images are supported. wl-clipboard's data-control protocol lets this agent
+observe and replace the Hyprland selection without holding keyboard focus.
+"""
+
+from __future__ import annotations
+
+import base64
+import hashlib
+import json
+import os
+from pathlib import Path
+import select
+import shutil
+import subprocess
+import sys
+import time
+from typing import Any
+
+
+PORT = Path("/dev/virtio-ports/dev.tryomarchy.clipboard")
+TEXT_FORMAT = "text/plain;charset=utf-8"
+PNG_FORMAT = "image/png"
+FORMATS = (TEXT_FORMAT, PNG_FORMAT)
+# Payload bytes before base64. Large enough for screenshots and pasted
+# documents, small enough that a runaway selection cannot exhaust memory.
+MAX_PAYLOAD_BYTES = 16 * 1024 * 1024
+MAX_LINE_BYTES = MAX_PAYLOAD_BYTES * 4 // 3 + 4096
+TEXT_TYPES = ("text/plain;charset=utf-8", "text/plain", "UTF8_STRING", "TEXT", "STRING")
+
+
+def log(message: str) -> None:
+    print(f"omarchy-native-clipboard-bridge: {message}", file=sys.stderr, flush=True)
+
+
+def fingerprint(mime: str, payload: bytes) -> str:
+    digest = hashlib.sha256()
+    digest.update(mime.encode())
+    digest.update(b"\0")
+    digest.update(payload)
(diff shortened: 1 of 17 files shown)


macos/Sources/OmarchyVMHelper/NativeClipboardBridge.swift
```diff
@@ -0,0 +1,293 @@
...
+        descriptor = try NativeBridgeSocket.connectSecure(path: socketPath, label: "clipboard bridge")
+        self.pasteboard = pasteboard
+        observedChangeCount = pasteboard.changeCount
+    }
+
+    deinit {
+        stop()
+    }
+
+    func run() throws {
```
themartiano, line 229: Here one syscall is being made per byte, pls optimize so bytes are read in chunks

guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
```diff
@@ -0,0 +1,287 @@
...
+        mime = TEXT_FORMAT
+    elif PNG_FORMAT in offered:
+        requested = PNG_FORMAT
+        mime = PNG_FORMAT
+    else:
+        return None
+    content = subprocess.run(
+        ["wl-paste", "--no-newline", "--type", requested],
+        stdin=subprocess.DEVNULL,
+        stdout=subprocess.PIPE,
```
themartiano, line 191: Codex:
`subprocess.run(stdout=PIPE)` first loads an arbitrarily large selection, then base64 creates another allocation. A large clipboard can exhaust memory before `guest_changed()` rejects it. Read at most `MAX_PAYLOAD_BYTES + 1` from a pipe and terminate oversized transfers.

macos/Sources/OmarchyVMHelper/NativeClipboardBridge.swift
```diff
@@ -0,0 +1,293 @@
...
+            throw HelperError.io("guest sent an unknown clipboard message type")
+        }
+    }
+}
+
+/// Echo-safe two-way bookkeeping shared by the tests and the live bridge.
+/// Applying the guest's selection to the Mac pasteboard bumps the pasteboard
+/// change count, and writing the Mac pasteboard into the guest triggers the
+/// guest watcher; both are recognized by fingerprint and dropped.
+struct ClipboardSyncState: Equatable {
```
themartiano, line 109: The echo fingerprints never expire, they should otherwise we could lose data. Example: guest copies B, Mac copies A, then Mac copies B again; final B is incorrectly discarded because of the original echo

guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
themartiano, line 1: Codex:
QEMU delivers EOF when its chardev client disconnects, but returning 0 combined with `Restart=on-failure` (`omarchy-native-clipboard-bridge.service#L7-L10`) means the guest agent stays stopped. Consequently, the launcher’s first advertised host-bridge restart will not restore clipboard sharing. Reopen internally, exit nonzero, or use `Restart=always`

themartiano (commented):
This feature is very welcome, thank you!
It was in my todo list

macos/Sources/OmarchyVMHelper/NativeClipboardBridge.swift
rlimberger: Fixed in 600a316: `run()` now reads into a 64 KiB buffer and splits on newlines in the chunk; the per-line size guard is unchanged.

guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
rlimberger: Fixed in 600a316: `wl-paste` now runs under `Popen` and its stdout is drained by `read_bounded()`, which reads at most `MAX_PAYLOAD_BYTES + 1` (64 KiB at a time, with the old 15 s deadline) and kills the child as soon as the limit is passed. Nothing oversized reaches base64/JSON. Covered by the new `BoundedReadTests`.

macos/Sources/OmarchyVMHelper/NativeClipboardBridge.swift
rlimberger: Good catch — fixed in 600a316 on both sides. Each direction now stores (fingerprint, timestamp) and a marker only counts as an echo for `echoWindow` / `ECHO_WINDOW_SECONDS` (2 s). Your exact sequence (guest B → Mac A → Mac B) is the new test on both the Swift and Python state machines.

guest/native-overlay/usr/local/bin/omarchy-native-clipboard-bridge
rlimberger: Fixed in 600a316: EOF on the port now raises `RuntimeError("host clipboard bridge disconnected")` → exit 1, so `Restart=on-failure` brings the agent back and it re-sends `sync` once the launcher has restarted the host bridge. Added `StartLimitIntervalSec=0` to the unit so several host-side restarts in a row cannot trip systemd's start limit.

macos/Sources/OmarchyVMHelper/NativeClipboardBridge.swift
themartiano: Seems to still be happening.

Eg:
1. Guest copies B.
2. Mac copies A.
3. Mac copies B.
Step 2 does not clear the old guest-B marker, so step 3 is incorrectly discarded.

rlimberger: You're right — the expiry alone didn't cover it inside the window. Fixed in a39f541: accepting a change in one direction now clears the other direction's marker, so step 2 (Mac copies A) drops the guest-B marker and step 3 goes through. That's safe because a marker only protects content still on the other side's clipboard, and the echo of a write is always delivered before any later user copy in the same event stream. The 2 s expiry stays as a backstop. Your exact sequence, and its mirror image (Mac C → guest D → guest C), are now tests on both the Swift and Python state machines.

PR #9, opened 2026-08-26
Share one Mac folder with Omarchy over virtio-9p

### LINKED ISSUE #11: Share a Mac folder with the Omarchy guest
There is no way to move files between the Mac and Omarchy other than the network. A shared folder with read/write access from both sides would make the VM far more usable for real work.

**Expected:** pick one Mac folder in the start menu; it appears inside Omarchy under the same name (a shared `~/Work` shows up as `~/Work`) with full read/write access. Sharing stays off until a folder is chosen; the whole home folder, `~/Library`, and system directories are refused.

**Notes:** virtio-9p works on macOS QEMU with no extra host dependencies. The obstacle is ownership: the Mac user's uid differs from the Omarchy account's uid 1000, Linux 9p does not support idmapped mounts, and `bindfs` is not packaged for Arch Linux ARM, so the mapping has to happen on the host side (a small QEMU 9p patch).

Implemented in #9.

rlimberger:
Closes #11

## Summary

- The start menu gains a **Shared folder** row (Choose…, Turn On/Off). One Mac folder is exported to the guest over virtio-9p and linked into the Omarchy home under its own name, so a shared `~/Work` appears as `~/Work`.
- Ownership: Linux 9p has no idmapped-mount support and `bindfs` is not packaged for Arch Linux ARM, so the mapping lives on the host. A small QEMU patch (`macos/patches/qemu-9p-guest-owner.patch`, applies cleanly to the pinned commit) adds `guest_owner_uid`/`guest_owner_gid` fsdev options for `security_model=none`: host operations still run as the Mac user and the Mac keeps real modes/ownership, but that user's files are reported as the first Omarchy account (uid/gid 1000) so guest permission checks agree with the host. `--enable-virtfs` is added to the runtime build and the device/options are verified at build, prepare, and launch time.
- Policy: the Swift app validates and canonicalizes the choice (no `/`, `/Users`, the whole home, `~/Library`, symlinks, non-owned dirs, or paths containing `,`), persists it in UserDefaults, and passes it via `OMARCHY_QEMU_GPU_SHARED_FOLDER`; `run-qemu-gpu.sh` re-validates before emitting the fsdev/device pair. The folder name reaches the guest as `omarchy.shared_folder_name=<base64url>` on the kernel command line. `Info.plist` gains Desktop/Documents/Downloads usage strings for the TCC prompt.
- Guest: `omarchy-native-mac-share` mounts the `mac` tag at `/mnt/mac` before the display manager; a user unit links `~/<name>` at login, replacing only an empty xdg folder of the same name and falling back to `~/Mac` otherwise. The pinned kernel builds 9p as modules.
- Contract, validator, `verify.py`, and Swift tests cover the device, options, policy, preferences, and link behavior.

Sharing is off until a folder is chosen, so existing users see no change. A guest rebuild changes the factory image identity (reset required), as with any `guest/` change.

## Test plan

- [x] `swift test` (52 tests) and `guest/test` pass
- [x] `make runtime`: virtfs enabled, patch applies, `virtio-9p-pci` present; invalid option combinations are rejected with the intended messages
- [x] Full `make build`, first-boot provisioning, then browsing a shared Mac folder from the guest
- [ ] Files created in the guest show up owned by the Mac user on macOS (expected from security_model=none; not yet checked by hand)

Note: building on current Homebrew (sdl2-compat/SDL3) needs #6 first; this PR does not touch those pins. Independent of the clipboard PR; both touch `run-qemu-gpu.sh`, `spec.json`, `configure-rootfs.sh`, and `verify.py` in adjacent hunks, so whichever lands second will need a trivial rebase.

https://claude.ai/code/session_01GopAGVNHBeeDUwEQ21Qotf

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,9 @@ into its home under the same name (`~/Work` on the Mac becomes `~/Work` in
 Omarchy) with full read and write access, so choose a folder you intend Linux
 software to modify. The whole home folder, `~/Library`, and system directories
 cannot be shared. **Turn Off** keeps the choice but stops exporting it on the
-next launch. The share belongs to the first Omarchy account created during
+next launch; Omarchy then removes the link and gives back any standard folder
+such as `~/Documents` that the link had taken over. The share belongs to the
+first Omarchy account created during
 provisioning; additional guest accounts see it read-only.
 
 ## Requirements
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Try Omarchy is not official or affiliated with Omarchy.
 - Resizable native window with automatic guest resolution and HiDPI scale updates
 - Mac audio input/output selection inside Omarchy, with live routing and system-default fallback
 - Two-way clipboard sharing for text and PNG images between macOS and Omarchy
+- One optional shared Mac folder, available inside Omarchy under the same name (`~/Work` stays `~/Work`)
 
 > **Current limitation:** Video decoding is CPU-only, so playback can be slow, especially at high resolutions. An improved video path is in development.
 
@@ -31,6 +32,17 @@ Copy and paste work in both directions as soon as you sign in to Omarchy: text
 and PNG images copied on the Mac appear in the Omarchy clipboard, and content
 copied in Omarchy lands on the Mac pasteboard. Nothing is transferred until
 something is copied.
+
+## Sharing a folder with the Mac
+
+Folder sharing is off until you pick a folder. Use **Choose…** next to
+**Shared folder** on the start menu to select one Mac folder; Omarchy links it
+into its home under the same name (`~/Work` on the Mac becomes `~/Work` in
+Omarchy) with full read and write access, so choose a folder you intend Linux
+software to modify. The whole home folder, `~/Library`, and system directories
+cannot be shared. **Turn Off** keeps the choice but stops exporting it on the
+next launch. The share belongs to the first Omarchy account created during
+provisioning; additional guest accounts see it read-only.
 
 ## Requirements
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -42,6 +42,16 @@ fingerprint of what it last wrote so the immediate echo is dropped. The marker
 is cleared as soon as the other side moves on to new content, and expires after
 a couple of seconds regardless, so a genuine repeat of the same content still flows.
 
+When a folder is chosen on the start menu, QEMU exports it over virtio-9p with
+`security_model=none`, so every host file operation runs as the Mac user and
+the Mac keeps real modes and ownership. A small QEMU patch adds
+`guest_owner_uid`/`guest_owner_gid` fsdev options that report the Mac user's
+files as the first Omarchy account (uid/gid 1000), which makes the guest
+kernel's permission checks agree with what the host will actually allow. The
+guest mounts the tag at `/mnt/mac` before the display manager starts, and a
+user unit links `~/<folder name>` to it at login; the name travels on the
+kernel command line as `omarchy.shared_folder_name=<base64url>`.
+
 ## The ARM64 image
 
 The guest image is built by this project; it is not an official prebuilt image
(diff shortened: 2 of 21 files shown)


themartiano:
Codex report:
- **[P1] Turning sharing off leaves a stale home-directory link.** The user service is skipped when `/mnt/mac` is not mounted, and the script also returns before removing previous links. After sharing `Documents`, for example, turning sharing off can leave `~/Documents` pointing at the now-unmounted, root-owned `/mnt/mac`, breaking normal document access. Cleanup should run even without an active mount and restore any displaced XDG directory. [[User unit](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/guest/native-overlay/usr/lib/systemd/user/omarchy-native-mac-share-link.service#L1-L7)](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/guest/native-overlay/usr/lib/systemd/user/omarchy-native-mac-share-link.service#L1-L7), [[early return before cleanup](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/guest/native-overlay/usr/local/bin/omarchy-native-mac-share#L88-L104)](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/guest/native-overlay/usr/local/bin/omarchy-native-mac-share#L88-L104)

- **[P2] Sharing disabled adds about five seconds to every boot.** The mount service is always enabled and ordered before the display manager, but an absent share still executes ten 500 ms sleeps. The missing kernel name parameter could be used to return immediately when sharing is off. [[Polling loop](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/guest/native-overlay/usr/local/bin/omarchy-native-mac-share#L39-L51)](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/guest/native-overlay/usr/local/bin/omarchy-native-mac-share#L39-L51)

- **[P2] The shell validator does not recheck delimiters after canonicalization.** The comma/newline check occurs before `pwd -P`. An intermediate symlink can therefore resolve a comma-free input to a comma-containing canonical path, which is then inserted into QEMU’s comma-delimited `-fsdev` options. A crafted environment value could make QEMU export a different directory, including one otherwise prohibited by policy. Re-run all character checks after line 636. [[Launcher validation](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/macos/run-qemu-gpu.sh#L628-L649)](https://github.com/themartiano/try-omarchy/blob/938ef235e6957d578a566f087d3465c54ac0ad54/macos/run-qemu-gpu.sh#L628-L649)

rlimberger:
Thanks — all three addressed in a649841:

- **P1 stale link:** `--link` now runs at every login regardless of the mount (dropped `ConditionPathIsMountPoint`). Without a mount it removes every link pointing at `/mnt/mac` and, if the link had replaced an xdg folder (read from `user-dirs.dirs`, falling back to the stock names), recreates that folder. Covered in `verify.py`.
- **P2 boot delay:** `--mount` returns immediately when `omarchy.shared_folder_name` is absent from the kernel command line; the 10×500 ms poll only runs when the launcher actually exported a share. There is also a `--enabled` query for scripting/tests.
- **P2 validator:** the absolute-path / plain-directory / newline / CR / comma checks are re-run on the canonical path right after `pwd -P`, before it can reach `-fsdev`. Verified with a symlink whose target directory has a comma in its name.

themartiano:
Pls merge main to fix the CI, other than that LGTM 🚀

rlimberger:
Rebased onto main (now including #8) and resolved the overlaps in the docs, spec, `configure-rootfs.sh`, `verify.py`, and the launcher — both features are kept side by side. Also updated the new start-menu layout test from #15 to pass the shared-folder callbacks, which is what CI was failing on. `make test` passes locally (Swift suite, guest contract, storage script).

PR #35, opened 2026-08-27
Fix launch when GNU stat shadows macOS stat

stefankuksenko:
## Summary

- use `/usr/bin/stat` for BSD-specific filesystem metadata checks
- reuse the storage helpers for launcher ownership and permission checks
- run the storage regression suite with a deliberately failing `stat` first in `PATH`

## Problem

When Homebrew coreutils precedes the system tools in `PATH`, the launcher resolves GNU `stat` for BSD-only invocations such as `stat -f '%u'`. GNU `stat` interprets those arguments differently:

```
stat: cannot read file system information for '%u': No such file or directory
run-qemu-gpu: temporary directory is not owned by this user
```

The app then reports only that its virtual machine stopped during startup. Pinning these security-sensitive metadata checks to macOS `/usr/bin/stat` avoids the PATH-dependent behavior without hiding Homebrew Python or other development tools.

## Testing

- `make test`
- manually reproduced the failure with Homebrew GNU coreutils first in `PATH`
- verified the storage lifecycle test passes with a failing shadow `stat` first in `PATH`

The staged-QEMU lock inheritance subtest was skipped because this checkout has no locally built QEMU runtime; all other tests passed.

### DIFF
--- a/macos/Tests/qemu-persistent-storage.test.sh
+++ b/macos/Tests/qemu-persistent-storage.test.sh
@@ -130,18 +130,30 @@ cleanup() {
 }
 trap cleanup EXIT HUP INT TERM
 
+# Runtime filesystem checks must use macOS BSD stat even when Homebrew
+# coreutils shadows it in the caller's PATH.
+shadow_bin="$test_root/path-shadow"
+mkdir "$shadow_bin"
+printf '%s\n' \
+  '#!/bin/bash' \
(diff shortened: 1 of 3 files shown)


themartiano:
Great! Thanks for the fix.
Please update your branch and I'll merge

stefankuksenko:
done

PR #38, opened 2026-08-28
Add opt-in localhost SSH access

CN-Scars:
## Summary

This PR adds a focused SSH preset to Try Omarchy's existing generic Port Forwarding UI.

Choosing **Add SSH** creates an ordinary, editable mapping:

```text
TCP 127.0.0.1:2222 -> guest:22
```

After saving the mapping and launching Omarchy, users can connect with an explicit guest username:

```bash
ssh -p 2222 <guest-user>@127.0.0.1
```

The mapping remains owned by the generic port-forwarding model. This PR does not add a second SSH-specific preference, port probe, bind-error path, or QEMU `hostfwd` builder.

Try Omarchy starts with no mappings, and every host forward remains bound to IPv4 loopback. Nothing is exposed on Wi-Fi, Ethernet, `0.0.0.0`, or another LAN-visible address.

## User experience

With no mappings configured, the start menu shows Port Forwarding as **Off**. This is the default state for new users.

From the start menu:

1. Open **Port Forwarding** with **Configure…**.
2. Choose **Add SSH**.
3. Review or edit the ordinary `TCP 2222 -> 22` mapping.
4. Save and launch Omarchy.

The preset does not auto-save, and the generated row remains fully editable. It uses the generic editor's existing validation, 32-mapping limit, duplicate `(protocol, hostPort)` policy, persistence, port-availability checks, and startup error recovery.

After saving, the start menu reports the effective generic mapping, for example:

```text
localhost:2222 -> Omarchy:22 · TCP
```

There is intentionally no **Copy Command** action: Try Omarchy does not know or guess the guest account name. Connection examples use `<guest-user>` explicitly.

### Generic host configuration

The preset is a normal `PortForwardMapping` with:

```text
protocol = tcp
hostPort = 2222
guestPort = 22
```

It is persisted and exported through the generic port-forwarding path:

```text
PortForwardMapping
  -> OMARCHY_QEMU_GPU_PORT_FORWARDS
  -> qemu-port-forwarding.sh
  -> QEMU SLIRP hostfwd
```

The existing policy continues to support arbitrary TCP and UDP mappings with canonical ports in `1...65535`. An occupied or duplicate TCP host port produces the existing actionable generic error instead of being silently reassigned.

### QEMU networking and SSH boot intent

With no mappings, the network argument remains:

```text
-netdev user,id=omarchy-net
```

A saved SSH preset produces:

```text
-netdev user,id=omarchy-net,hostfwd=tcp:127.0.0.1:2222-:22
```

The sealed shell parser independently validates the complete generic mapping list and constructs every fixed loopback-only `hostfwd` clause. Only after the entire list succeeds does it derive SSH boot intent from a valid TCP mapping to guest port 22.

This applies to any valid host port, not only the preset:

```text
tcp:<host-port>:22 -> tryomarchy.ssh_access=1
```

UDP mappings to guest port 22 and unrelated TCP mappings do not activate sshd. Malformed configuration publishes neither partial forwarding arguments nor stale SSH intent.

### Guest activation

OpenSSH is an explicit, locked factory package. A systemd generator recognizes the exact kernel token:

```text
tryomarchy.ssh_access=1
```

It then adds a runtime-only `multi-user.target.wants/sshd.service` link. The effect is boot-scoped:

- a TCP mapping to guest port 22 requests sshd for normal and ephemeral launches;
- removing all TCP guest-port-22 mappings stops Try Omarchy from requesting sshd on the next boot;
- reset-only mode never launches QEMU or activates sshd;
- no persistent sshd enable link is written under `/etc`.

Try Omarchy does not modify `sshd_config`, password policy, guest accounts, authorized keys, or host-key verification. It owns host reachability and boot activation only; authentication policy remains owned by Omarchy and the guest user.

## SSH onboarding

Complete Omarchy's account provisioning before connecting. For the initial connection, use the password created for that guest account:

```bash
ssh -p 2222 <guest-user>@127.0.0.1
```

Users can then install a public key:

```bash
ssh-copy-id -p 2222 <guest-user>@127.0.0.1
```

An optional `~/.ssh/config` entry provides a short alias:

```sshconfig
Host omarchy
  HostName 127.0.0.1
  Port 2222
  User <guest-user>
```

After that, `ssh omarchy` works naturally with `scp`, `rsync`, Git, and VS Code Remote SSH.

## Security properties

- Opt-in: the default generic mapping list is empty.
- Every host forward uses `127.0.0.1`; there is no wildcard, LAN, bridged, or vmnet exposure.
- Loopback prevents other devices on the current network from connecting, but it is not private to the current macOS account: other local users or processes can still attempt authentication.
- The shell boundary rejects malformed, oversized, noncanonical, out-of-range, and duplicate mappings before VM storage or QEMU is touched.
- A launcher-owned `tryomarchy.ssh_access=` token is rejected if pre-baked into the factory command line.
- Guest authentication and OpenSSH host-key verification remain required and unchanged.
- The factory image contains no shared or pre-generated SSH host private keys.

## Host-key lifecycle

A persistent VM keeps its generated SSH host identity across ordinary restarts.

Factory Reset creates a new guest disk and therefore a new SSH host key. Every ephemeral VM also has its own host identity. If an endpoint is reused after Reset or for another ephemeral instance, remove only that endpoint's stale entry before reconnecting:

```bash
ssh-keygen -R '[127.0.0.1]:2222'
```

Substitute the configured host port when the preset was edited. The implementation does not disable strict host-key checking, delete entries automatically, or bake a shared private key into the factory image.

## Persistent VM compatibility

The saved VM remains tied to the exact immutable factory identity.

Because the explicit OpenSSH request and generator change that identity, an existing VM created by a different factory build follows Try Omarchy's existing confirmed **Reset Omarchy** flow. The implementation does not relabel incompatible state or silently overwrite a saved disk.

Generic forwarding preferences are host-side state and survive app restarts and Factory Reset. A compatible persistent VM is reused when mappings change; forwarding alone does not rebuild it.

## Scope

This is deliberately a small SSH preset on top of the already-merged generic Port Forwarding feature.

It does not add:

- an SSH-specific preference schema or second forwarding pipeline;
- guest username discovery or persistence;
- automatic host-port reassignment;
- LAN, wildcard, bridged, or vmnet networking;
- guest IP discovery;
- automatic user SSH-key generation or installation;
- changes to Omarchy's SSH authentication policy;
- automatic host-key deletion or host-key verification bypasses.

### Automated

The full `make test` suite passes on the rebased tree, including:

- build-cache and guest/update tests;
- the strict native guest verifier;
- the complete Swift test suite;
- generic port-forwarding policy, editor, availability, persistence, and shell contracts;
- the focused SSH launcher contract;
- persistent-storage, reset, compatibility, relocation, and power-action suites;
- the upstream storage-location and camera-related coverage.

The SSH-specific coverage includes:

- exact preset insertion as an ordinary editable `tcp:2222:22` mapping;
- duplicate TCP host-port feedback, UDP coexistence, accessibility, Save behavior, and the 32-rule limit;
- disabled, unrelated, UDP/22, default and nondefault TCP/22, and mixed mappings;
- all-or-nothing malformed-input handling and repeated-call state reset;
- loopback-only QEMU arguments and exactly one boot token;
- normal, ephemeral, reset-only, and persistent-disk reuse behavior;
- exact-token systemd generator behavior and runtime-only output;
- explicit and locked OpenSSH packaging;
- rejection of a pre-baked launcher-owned token.

### Build and live Apple Silicon validation

- `make app`: **pass**
- self-contained QEMU runtime relocation and macOS 15 compatibility validation: **pass**
- locally built ad-hoc app strict/deep signature verification: **pass**
- factory ext4 inspection for pre-generated `ssh_host_*` private keys: **pass** (none present)
- real QEMU/HVF guest launch and Omarchy provisioning: **pass**
- password-authenticated host-to-guest SSH through SLIRP: **pass**
- normal guest reboot followed by automatic SSH recovery: **pass**
- strict host-key reuse across the persistent restart: **pass**

Observed host listener:

```text
TCP 127.0.0.1:2222 (LISTEN)
```

No wildcard, `0.0.0.0`, or IPv6-any listener was present on the Mac.

Observed guest state:

```text
architecture:         aarch64
sshd.service:         active
enablement:           enabled-runtime
boot token count:     1
persistent /etc link: absent
```

`enabled-runtime` is the expected systemd-generator result; it does not persistently enable sshd under `/etc`.

This local ad-hoc build does not claim Developer ID, notarization, stapling, quarantine, or clean-user Gatekeeper release qualification.

## Documentation

This PR updates the top-level, guest, macOS, architecture, and release documentation to cover:

- the generic **Add SSH** workflow;
- explicit guest usernames and initial password authentication;
- public-key installation and SSH aliases;
- `scp`, `rsync`, Git, and VS Code Remote SSH usage;
- localhost's real security boundary;
- boot-scoped guest activation ownership;
- persistent VM, Factory Reset, and ephemeral host-key behavior;
- endpoint-specific known-host recovery;
- release validation expectations.


themartiano:
Very good! This has been requested a couple of times already.

Codex:
* The README says SSH supports factory resets and ephemeral launches, but Arch generates missing SSH host keys inside each guest disk. A reset creates a new disk, and each ephemeral launch starts from the factory image, so the host key changes while the endpoint remains `[127.0.0.1]:2222`. OpenSSH will then reject the connection with “REMOTE HOST IDENTIFICATION HAS CHANGED.” See the [claimed lifecycle behavior](https://github.com/themartiano/try-omarchy/blob/e9a9ad061069548501d3ca8b93f130e6e08888c0/README.md#L68-L72) and [guest activation](https://github.com/themartiano/try-omarchy/blob/e9a9ad061069548501d3ca8b93f130e6e08888c0/guest/native-overlay/usr/lib/systemd/system-generators/try-omarchy-ssh-access#L23-L31). Arch documents that `sshdgenkeys` generates missing host keys automatically: [ArchWiki](https://wiki.archlinux.org/title/OpenSSH), [package contents](https://archlinux.org/packages/core/x86_64/openssh/files/).

* "Copy Command" may use the wrong username. The copied command assumes the Mac and Omarchy usernames match. If they differ, SSH fails. Ask for the Omarchy username or include it in the saved SSH settings.

* Security wording should clarify the localhost boundary. Loopback prevents LAN exposure, which is excellent, but it is not private to the current Mac account. Other local processes/users can attempt authentication against the listener.


What do you think of this?
* Authentication onboarding is too thin, add:
   - Complete Omarchy account provisioning before connecting.
   - Initial password-authentication expectations.
   - A concrete `ssh-copy-id -p 2222 guest@127.0.0.1` example.
   - An `~/.ssh/config` entry so `scp`, `rsync`, Git and VS Code Remote SSH work naturally.
   - Host-key recovery instructions for resets, unless handled automatically.

CN-Scars:
Thanks — I agree with all of these points. I’ll expand the authentication onboarding, clarify the localhost security boundary, and document host-key recovery after resets/ephemeral launches rather than weakening host-key verification.

I also noticed that #44 has now landed and overlaps with this PR’s SSH-specific port preferences, availability checks, UI, and QEMU `hostfwd` construction. I don’t want to keep two competing forwarding implementations.

Before I rebase and resolve the conflicts, which direction would you prefer?

1. Add an SSH preset to the new generic Port Forwarding UI (`TCP 2222 → guest 22`), keep the boot-scoped sshd activation, and put the username/connection setup in the onboarding docs.
2. Keep a dedicated SSH row, but rebuild it on top of the generic forwarding engine and save the guest username for Copy Command.

My preference is option 1 because it keeps one source of truth and makes this PR much smaller, but I’ll wait for your direction before reworking it.

themartiano:
Sounds good, let's go with 1. Thanks

CN-Scars:
Thanks! I’ve updated the PR based on your suggestion. SSH is now an Add SSH preset within the generic Port Forwarding editor, creating an ordinary and fully editable TCP localhost:2222 → Omarchy:22 mapping instead of using a dedicated SSH toggle or configuration path. I’ve also rebased the branch, updated the tests, documentation, PR description, and screenshots, and verified real SSH connectivity—including reconnecting successfully after a guest reboot.

themartiano:
Great work! Merging with CI failures because they come from the brittle UI tests, gotta remove those

PR #40, opened 2026-08-28
Add an opt-in QEMU guest agent channel

ramarivera:
Try Omarchy currently has no durable host-to-guest command channel. A live QMP experiment proved that the standard QEMU Guest Agent works in the ARM64 guest, including synchronized protocol calls and a harmless `guest-exec`, but the hot-plugged channel disappeared on restart.

This change makes that proven path reproducible without exposing it in normal releases:

- pins `qemu-guest-agent` in the reviewed ARM64 package transaction
- declares and validates the standard `org.qemu.guest_agent.0` runtime contract
- adds `OMARCHY_QEMU_GUEST_AGENT=1` as an explicit developer-only switch
- creates the socket inside the existing mode-0700 per-boot runtime directory and assigns the next free virtio-serial port
- leaves the root-capable channel absent when the switch is unset

The full repository test suite passes: guest contracts, Swift tests, persistent-storage integration, and the new QGA contract test. The regression test was also proven against a sabotaged port assignment, where it failed before the correct assignment was restored.

The package lock refresh also picked up the current ARM repository's HarfBuzz update and `numactl`, a QGA transaction dependency.

Security note: QGA can execute guest commands as root. This PR does not add a network listener, UI toggle, or release-default access path.

Assisted by GPT-5.6 Luna through OMP and GPT-5.6 Sol through Hermes Agent.

🤖 *This content was generated with AI assistance using GPT-5.6 Luna and GPT-5.6 Sol.*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ test:
 	@mkdir -p $(ROOT)/macos/.build/module-cache/swift $(ROOT)/macos/.build/module-cache/clang
 	@cd $(ROOT)/macos && SWIFT_MODULECACHE_PATH=$(ROOT)/macos/.build/module-cache/swift CLANG_MODULE_CACHE_PATH=$(ROOT)/macos/.build/module-cache/clang swift test --disable-sandbox
 	@$(ROOT)/macos/Tests/qemu-persistent-storage.test.sh
+	@$(ROOT)/macos/Tests/qemu-guest-agent.test.sh
 
 guest:
 	@OMARCHY_FORCE_BUILD="$(FORCE)" "$(BUILD_CACHE)" \
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -42,6 +42,12 @@ fingerprint of what it last wrote so the immediate echo is dropped. The marker
 is cleared as soon as the other side moves on to new content, and expires after
 a couple of seconds regardless, so a genuine repeat of the same content still flows.
 
+Developer launches can set `OMARCHY_QEMU_GUEST_AGENT=1` to add a second private
+Unix-socket channel named `org.qemu.guest_agent.0`. The guest package is present,
+and systemd starts its static service when that virtio port appears. This channel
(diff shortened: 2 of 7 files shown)


themartiano:
Hey @ramarivera, should the PR move on or are you tackling those Copilot comments?

ramarivera:
👋🏻 I got sidetracked trying to improve the arm package situation but its not great news :(I will address comments later

themartiano:
Yeah, it's one of the hard problems with ARM compatibility. The Omarchy team is actively working on building AArch64 versions of the packages FYI

ramarivera:
> Yeah, it's one of the hard problems with ARM compatibility. The Omarchy team is actively working on building AArch64 versions of the packages FYI

Ah got it, I'm surprised they are doing it tho but not complaining lol .
Then with this new info do you think it makes sense for me to raise a pr trying to knock off some low hanging fruits, so you can tell me if it's the right way and then from there I just try to kill as many as we can without Omarchy team?

themartiano:
> > Yeah, it's one of the hard problems with ARM compatibility. The Omarchy team is actively working on building AArch64 versions of the packages FYI
> 
> Ah got it, I'm surprised they are doing it tho but not complaining lol . Then with this new info do you think it makes sense for me to raise a pr trying to knock off some low hanging fruits, so you can tell me if it's the right way and then from there I just try to kill as many as we can without Omarchy team?

Package-wise we can hold back a bit, should be mostly fixed this week. But about this QEMU guest channel what is the purpose of having it?

themartiano:

This adds a root-capable channel to every guest, but there is still no host consumer or documented way to discover the live socket. The new test also checks source text only, so it does not prove the channel is absent by default or that the enabled socket is private. Please document the concrete workflow, add an executed launcher contract for both states, and rebase the branch onto `main`.

themartiano (commented):
Hey, should we move forward with this PR or close it?

themartiano:
Closing for lack of activity

PR #42, opened 2026-08-28
Support VM reboot without closing the app

themartiano:
## Summary

- Configure QEMU to reset on guest reboot without exiting the app
- Configure guest shutdown to power off QEMU and close the app
- Add coverage for the QEMU power-action policy
- Document the resulting restart and shutdown behavior

## Testing

- Not run (not requested)


themartiano:
beautiful number

PR #46, opened 2026-08-28
Keep the Cocoa full grab while the VM window is key

caius72:
## Problem

Guest Super bindings work right after launch and then stop, until you click back
into the window — after which they work again for a while. The cause is in the
pinned QEMU's `ui/cocoa.m`, not in anything on our side.

`full-grab` installs a global `CGEventTap`, but the tap is only allowed to swallow
an event while the **mouse** is grabbed:

```objc
if ([view isMouseGrabbed] && [view handleEvent:event]) {   // cocoa.m:344
```

and `handleEventLocked` hands every Command chord back to macOS on the same
condition (cocoa.m:999, :1043 post-patch). We run `virtio-tablet-pci`, so
`notifyMouseModeChange` releases the grab the moment the guest driver binds
(cocoa.m:1218-1220), and `mouseExited` releases it again whenever the pointer
leaves the view. From then on host Command chords leak back to macOS mid-session.

## Change

Adds `QemuCocoaView -isKeyboardCaptured` and gates the tap and both Command
bail-outs on it:

```objc
return isMouseGrabbed || (eventsTap != NULL && [[self window] isKeyWindow]);
```

Keyboard capture follows our key window. The mouse grab keeps its upstream
meaning for cursor association and window titling, and a build without
`full-grab` is unaffected — `eventsTap` is NULL and this collapses to exactly
`isMouseGrabbed`.

Also corrects the comment at `macos/run-qemu-gpu.sh:949`, which claimed full grab
captures chords "before Spotlight ... can consume them". It does not: macOS
dispatches symbolic hot keys (Command-Space, Command-Tab) above every event tap,
so those still reach the host — which is also what keeps Command-Tab available as
a way out of a fullscreen guest.

## Verification

- Patch applies with zero fuzz on top of the four existing patches (texture-borrowing,
  gpu-spike, product-identity, dynamic-display) against pinned QEMU
  `cf3e71d8fc8ba681266759bb6cb2e45a45983e3e`.
- Built through `make runtime` / `make app`; the `isKeyboardCaptured` selector is
  present in the bundled binary, and the app has been running locally.
- Wired into `build-qemu-gpu-runtime.sh` the same way as the other four patches:
  path variable, existence check, sha256 pin, `verify_file_sha`, `patch -p1`.

Manual check: with the guest desktop up, move the pointer out of the window and
back **without clicking**, then press a Super chord. Before this change it leaked
to macOS; now it stays in the guest.


themartiano:
Nice, this will improve the UX. Thank you.

Please merge main, it includes the new "Immersive mode" toggle, which handles fullscreen and window behavior.
A simple merge won't do it, some changes are required so both continue to work.

Codex:

The new predicate handles the chord’s ordinary key-down/up events, but the preceding `NSEventTypeFlagsChanged` cases for left/right Command remain gated by `isMouseGrabbed`. Once `virtio-tablet` releases the mouse, Command-down is swallowed by the event tap without sending guest Meta; the following key reaches the guest unmodified. The primary bug therefore remains. Update both Command modifier cases to use the keyboard-capture predicate too. [Patch lines 40–52](https://github.com/themartiano/try-omarchy/blob/a132470884c87405dfc733b7b88a4f1561091630/macos/patches/qemu-cocoa-full-grab-focus.patch#L40-L52)

themartiano:
Closing because this was part of #62

PR #48, opened 2026-08-28
Let users choose where the VM's data lives

pantulis:
## Summary

- The start menu can now move the VM's persistent workspace to any
  folder the user picks, including one on an external drive, instead
  of always living under `~/Library/Application Support/Try Omarchy`.
- A chosen folder is used **exactly as given** — Omarchy never creates
  a folder inside it. It must already be empty, or already be a
  workspace Omarchy has used before; anything else (a folder with
  other files in it, or a drive's top level) is refused with a
  specific, actionable message rather than silently restructured.
- The drive must be local APFS: the storage library clones the factory
  image with `cp -c` and expands the working disk sparsely, which only
  APFS supports here — on exFAT/FAT/NTFS the same disk would claim its
  full size immediately. Network volumes are refused because the disk
  lock is unreliable on them. Both the Swift app (at pick time) and the
  shell storage library (again at launch) validate independently,
  since the volume can change in between.
- Ejecting the drive mid-run is detected and shuts the VM down instead
  of leaving it writing to a vanished mount.
- Resetting a custom-location VM now composes the same environment as
  a normal launch, so it reliably erases the workspace actually in
  use rather than the default one (a latent bug this change would
  otherwise have newly exposed).
- A pre-existing folder with the "wrong" permissions (e.g. `0755` from
  Finder's New Folder) is corrected in place rather than rejected.

## Test plan

- [x] `swift test` — 95 tests across 21 suites, including new coverage
      for: empty-folder acceptance, Finder-cruft-is-ignored,
      non-empty-folder rejection, volume-root rejection,
      space-floor/comfort-target math, and launch/menu-state wiring
- [x] `macos/Tests/qemu-persistent-storage.test.sh` — including new
      cases for unsupported filesystem and insufficient free space
- [x] `guest/test` — unaffected, full contract still passes
- [x] Manual, on real hardware (external APFS drive):
  - [x] Default location unaffected
  - [x] Pick an empty folder on an external APFS drive → used
        directly, `disks/`/`images/`/`locks/`/marker land flat inside
        it, no nested subfolder
  - [x] Pick a folder with unrelated content already in it → rejected
  - [x] Pick a drive's top level → rejected
  - [x] Pick an exFAT-formatted volume → rejected, correct format
        named
  - [x] Pick an APFS volume too small for the factory image → rejected
        with the actual needed/available figures
  - [x] Unplug the drive mid-run → VM shuts down and a dialog explains
        why, instead of hanging or continuing to write
  - [x] Reset while on a custom location → erases that workspace,
        confirmed the default location is untouched
  - [x] Switch back to the default location and relaunch → the
        previously-used external-drive VM is untouched and reachable
        again by switching back
  - [x] A Finder-created folder (mode `0755`) is accepted and
        corrected to `0700` rather than rejected

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_014LHCLwYmLr6YrJh9mKA4Pz


themartiano:
Great addition! I'm all for flexibility

Codex findings:
* Reset can silently fall back from an unavailable custom drive to the default VM, erasing the wrong workspace.
* Opening a missing custom location can recreate it on the wrong filesystem and initialize a new VM there.
* The folder-picker text incorrectly says Omarchy creates a nested folder; it actually uses the selected folder directly.
* The folder picker treats any folder containing a file named .omarchy-qemu-storage as an existing Omarchy workspace, without checking that the marker file is valid. This bypasses the empty-folder safety check and can make the UI accept a folder that later fails to launch.

pantulis:
Thanks! All four Codex findings are fixed in 715db51. Each was real; details below.

**1. Reset erasing the wrong workspace.** Confirmed and traced. `resetVirtualMachine` never gated on the chosen folder being reachable the way `startVirtualMachine` does, and `StorageLocationLaunchConfiguration.make` answered a failed validation by logging one line to stderr — invisible in a GUI app — and omitting `OMARCHY_QEMU_GPU_STATE_ROOT`. The launcher fell back to the default workspace, erased `disks/current`, and since the key is `current`, `_qps_reset_remaining_legacy_workspaces` swept every sibling workspace beside it. Exit 0, so the app reported **success** while the VM the user meant to reset sat untouched on the unplugged drive. The confirmation sheet also stated *"The VM being erased is the one stored at /Volumes/…"*, which was false.

Fixed in two layers. The fallback is now explicit — `unavailableReason` travels with the launch configuration and through `ChildLaunchContext`, and both `launch` and `resetVirtualMachine` refuse while it is set, so no future caller can reintroduce a silent fallback. Above that, availability reports what the user chose rather than a bare yes/no: switching to the default is a fine way to *start* a VM so launch proceeds, but it must never be a way to *erase* one, so reset abandons the run and the user gets an accurate confirmation next time they ask. `resetDidAbort` unwinds the menu without claiming anything was reset, and the sheet no longer names a workspace it cannot reach.

**2. Opening a missing location rebuilding it on the wrong filesystem.** The old check — "does the parent exist?" — is satisfied by a leftover `/Volumes/<name>` directory on the boot volume, so the workspace was recreated on the internal disk and the next launch initialized a new VM there. Only the default folder is created from the reveal action now; a chosen folder that has gone missing means its drive is absent, and nothing is written. The row also drops the click affordance entirely once it sees a problem, so the guard is the backstop for the case where you click before the window redraws.

**3. Picker text.** One-line copy fix — it had described the old nesting behavior. Now: *"Omarchy puts its VM files straight into the folder you choose — it does not create a folder inside it. Pick an empty folder, or one Omarchy already uses. The drive must be APFS."*

**4. Marker trusted on sight.** `_qps_validate_root_marker` requires a 0600 regular file owned by the caller containing exactly `omarchy-qemu-storage-root-v1`; Swift only checked the name existed. An empty file, a directory, a symlink, or a 0644 copy restored from a backup was accepted, skipped **both** the volume-root and emptiness guards (the marker name is itself in `ignorableEntries`), and failed later with a message from the shell script rather than the picker. `hasValidRootMarker` mirrors the launcher exactly, and a damaged marker is refused by name rather than falling through to a misleading "folder isn't empty".

Worth flagging: the existing fixtures wrote an **empty** marker, so the tests encoded the bug. They now write a real one, and `StorageLocationContractTests` pins the marker filename and token against `qemu-persistent-storage.sh` — the shell hardcodes the filename inline with no constant, so nothing previously kept the two languages in sync.

**Verification.** 132 Swift tests and the full `qemu-persistent-storage.test.sh` suite pass; the latter gained coverage for `_qps_validate_root_marker`, which had none. Manually verified against a real external drive: with it ejected, Reset refuses and the default workspace's factory image is byte-for-byte untouched (same mtime and size before and after), choosing *Use Default Folder* switches the setting without erasing anything, the missing folder is not recreated, and a folder carrying a bogus marker is rejected in the picker instead of at launch.

pantulis:
Rebased onto current `main` (through #53), so this now includes #47 and #49 rather than carrying them locally. CI is green.

Two CI failures along the way, both worth recording since neither reproduced on my machine:

**The start menu no longer fit its window.** Adding the VM Location row took the content to 717pt inside a 690pt window, leaving the launch button 15pt above the bottom edge — fine locally, clipped off the window under the runner's font metrics. The menu is now sized to 760 so it fits outright. The layout test also paired `canResetStorage: false` with a non-nil `storageLocation`, which the controller cannot produce (it returns no location for a disposable VM), so it was measuring a six-row layout users never see; the fixture now matches the real wiring.

**My new marker test called `stat` directly**, which the suite's PATH shim exists to catch — the same shadowing problem as #35. Now uses `/usr/bin/stat`, consistent with the rest of the file.

`make test` passes end to end locally against this base.

themartiano:
Awesome!

Codex found two small issues 😅
They're small but worth fixing IMO:

- **Reset can still erase a different workspace when the environment override is set.** The reset confirmation derives its location from the stored preference ([StartMenuWindow.swift](https://github.com/themartiano/try-omarchy/blob/b49fa1efde14e23e234a229113f9d6f89c3a14f0/macos/Sources/OmarchyVMHelper/StartMenuWindow.swift#L1124-L1140)), but the actual launch configuration deliberately lets `OMARCHY_QEMU_GPU_STATE_ROOT` override that preference ([StorageLocation.swift](https://github.com/themartiano/try-omarchy/blob/b49fa1efde14e23e234a229113f9d6f89c3a14f0/macos/Sources/OmarchyVMHelper/StorageLocation.swift#L382-L399)). Consequently, with preference A and override B, the confirmation says A while reset erases B. This is mainly a developer/test launch scenario, but the override is explicitly documented and retained, and the consequence is data loss. The confirmation and availability checks should use the same effective launch configuration as reset.

- **Swift and shell still disagree about valid marker contents.** Swift uses `trimmingCharacters(in: .newlines)`, which strips leading and trailing newlines ([StorageLocation.swift](https://github.com/themartiano/try-omarchy/blob/b49fa1efde14e23e234a229113f9d6f89c3a14f0/macos/Sources/OmarchyVMHelper/StorageLocation.swift#L194-L205)); shell command substitution strips only trailing linefeeds ([qemu-persistent-storage.sh](https://github.com/themartiano/try-omarchy/blob/b49fa1efde14e23e234a229113f9d6f89c3a14f0/macos/qemu-persistent-storage.sh#L244-L251)). A marker beginning with a newline is therefore accepted by the picker but rejected during launch—the exact class of mismatch the latest fix intended to eliminate. Swift’s `0777` mask also ignores special permission bits while the shell requires exactly `600`.

pantulis:
Both fixed in 3b6bcd3. One sub-point doesn't reproduce — details and evidence below.

**Reset under the environment override.** Confirmed and fixed. `StorageLocationMenuState` and the availability gate read only the stored preference, while `StorageLocationLaunchConfiguration` deliberately lets `OMARCHY_QEMU_GPU_STATE_ROOT` win, so with preference A and override B the sheet named A and the launcher erased B.

The menu state now resolves the override and reports it as the effective location, so the confirmation names the workspace reset will actually act on. The availability gate no longer validates a preference this run will not use — the launcher owns validating the override and fails loudly rather than falling back. The row also labels the override and disables `Change…`/`Use Default`, which would otherwise look like they work while changing nothing.

Rather than assert the two agree, `StorageLocationEffectiveConfigurationTests` now checks it across every preference × override combination; it fails against the previous menu state with `default, override set: menu shows default but the launcher uses /Volumes/Scratch/override-root`.

**Marker newline handling.** Confirmed and fixed — you're right, and my "mirrors the shell exactly" comment was wrong. `trimmingCharacters(in: .newlines)` strips both ends; `$(<file)` strips trailing linefeeds only. It now strips trailing linefeeds and nothing else. The same trim also swallowed a trailing `\r`, so a CRLF marker had the identical split; that case is covered too.

**Permission bits — this one I can't reproduce.** `_qps_permissions` is `/usr/bin/stat -f '%Lp'`, and `%Lp` reports only the nine permission bits, so the shell does not in fact require the special bits to be clear:

```
chmod 600   -> %Lp=600
chmod 4600  -> %Lp=600
chmod 2600  -> %Lp=600
chmod 1600  -> %Lp=600
```

`st_mode & 0o777` gives `0o600` for all four as well, so both sides accept a setuid marker today — equivalent, not laxer. I've noted the reasoning where the mask is applied instead of leaving it to be re-derived. Happy to tighten *both* sides to reject special bits if you'd prefer that; it needs `%OLp` (or `%p`) on the shell side, so it's a launcher contract change rather than a picker-only one.

**On the underlying pattern.** Both of these were semantic drift that grepping the shell for constants could never catch — it pins spelling, not behavior. `StorageMarkerAgreementTests` now executes *both* implementations over nineteen marker shapes (newline placement, CRLF, permission and special bits, wrong tokens, a token prefix, a directory, a symlink) and requires identical verdicts, sourcing `qemu-persistent-storage.sh` and calling `_qps_validate_root_marker` directly. Against the previous trim it fails on `leading newline`, `leading and trailing`, and `trailing carriage return`. That is the check that should stop this class of bug recurring.

`make test` passes locally and CI is green.

themartiano:
Great work! Thank you @pantulis

PR #58, opened 2026-08-29
Fix 1Password installation on ARM64 guests

dbachelder:
## Summary

- install the official signed 1Password ARM64 tarball when the guest is `aarch64`, while preserving Omarchy's package path on other architectures
- verify the vendor key bundle contains exactly the pinned primary signing key and verify the archive in an isolated GPG home, binding `VALIDSIG` to fingerprint `3FEF9748469ADBE15DA7CA80AC2D62742012EA22`
- install the ARM-capable `1password-cli` AUR package with the same isolated GPG home and non-interactive clean/diff choices
- set the application tree to `root:root` before running 1Password's vendor post-install script and install its missing `which` dependency
- route normal launches through a software-rendered wrapper because Electron cannot create its requested EGL context on Try Omarchy's virtual GPU
- add a Hyprland global binding for 1Password Quick Access and document both shortcuts:
  - `Ctrl + Shift + Space` — Quick Access
  - `Super + Shift + /` — full 1Password app
- declare the mutable post-build vendor/AUR trust boundary in the guest contract and third-party notices

## Testing

- `guest/test --source /tmp/try-omarchy-upstream.Woj9LX` — 38 unit tests and all guest contract checks pass
- `git diff --check`
- performed a clean uninstall and reinstall in an ARM64 Try Omarchy guest
- verified `/opt/1Password`, the application binary, Chrome sandbox, and wrapper have the required root ownership and modes
- signed in successfully and confirmed the full application opens with software rendering
- confirmed `Ctrl + Shift + Space` opens Quick Access through the Hyprland binding
- confirmed the normal post-install launch produced no new core dump

Full `make test` cannot run inside the Linux guest because the native targets require macOS and Swift tooling.

Related upstream context: basecamp/omarchy#87


themartiano:
Thanks! Lots of people will love this

PR #59, opened 2026-08-30
Release the SDL audio device while the guest is silent

caius72:
## Problem

QEMU holds a CoreAudio device open for the entire life of the VM, in both directions, even when the guest has never played a sound.

`sdl_enable_out`/`sdl_enable_in` only call `SDL_PauseAudioDevice`. QEMU links sdl2-compat over SDL3, where pausing a *logical* device leaves the *physical* device open with its audio thread still iterating — so the host keeps resampling silence.

Measured on an idle VM (Apple Silicon, guest at the desktop, nothing playing):

| | before | after |
|---|---|---|
| `coreaudiod` | **6.4%** of a core | **0.3%** |
| CoreAudio threads in the QEMU process | 5 | **0** |

The 0.3% figure is the machine's floor with no VM running at all.

The threads held were `com.apple.audio.IOThread.client` ×2, `SDLAudioP`, `SDLAudioC` and `AQConverterThread`.

Notably the device being held wasn't even from playback: `sdl_init_out` opens one at startup purely to negotiate a format, and nothing ever released it. A VM that never makes a sound still pays the full cost.

## Change

- Release the device when the frontend stops the stream, and after the initial format negotiation.
- Re-acquire it when the stream starts.
- Guard the route poller so it can't resurrect a released device.

The reopen reuses the spec the voice was initialised with, so the guest never observes a format change. It also re-reads the route file, so switching the Mac output device while the guest is silent continues to work.

## Guest side

No guest change needed — it was already behaving correctly. Every PipeWire sink reported `SUSPENDED` and `/proc/interrupts` showed `snd_hda_intel` at 0 interrupts/s. The host was burning CPU on a stream the guest had already stopped.

## Verification

- Patch applies to the pinned QEMU tree (`cf3e71d8`) with zero fuzz and no rejects.
- `audio_device_patch_sha256` updated in `build-qemu-gpu-runtime.sh` to match.
- Playback start/stop exercised repeatedly; audio resumes correctly after idle.

Not yet exercised on my side: switching the Mac output device *while the guest is silent*. The reopen path re-reads the route file so it should be picked up on the next acquire, but I haven't confirmed that end to end — worth a look in review.

Touches only `audio/sdlaudio.c`, so it's independent of the `ui/cocoa.m` and `util/cutils.c` patches in the stack.


themartiano:
Very good, thanks

PR #62, opened 2026-08-30
Move to QEMU 11.1.1 and Apple's in-hypervisor GIC

caius72:
Moves the runtime to QEMU 11.1.1 so the ARM GIC can live inside Hypervisor.framework instead of being emulated in QEMU userspace, and carries the patch stack forward to match.

## Why

QEMU emulated GICv2 in userspace under the Big QEMU Lock. Every guest interrupt cost roughly four lock acquisitions, and every IPI about five across two vCPU threads. The cost scaled with vCPU count and was independent of what the guest was doing, so an idle desktop still burned a core.

QEMU 11.1 is the first release carrying `hw/intc/arm_gicv3_hvf.c`, which hands the GIC to Apple's in-hypervisor implementation (`hv_gic_*`).

Measured A/B, headless, same disk, 90 s idle:

| | GICv2 (userspace) | GICv3 (`hv_gic`) |
|---|---|---|
| QEMU CPU time | 0.69 s | **0.08 s** |
| `hvf_wfi` exits | 34,444 | **0** |

At the desktop, idle went from ~65% of a core to ~22%.

Under HVF, QEMU 11.1.1 also rejects GICv2 outright, so `gic-version=3` is now the only supported configuration rather than an optimisation.

## What is in here

- **QEMU `cf3e71d8` (10.2.50) → `c3d48b7d` (11.1.1)**, `-machine virt,accel=hvf,gic-version=3`.
- **The render patch is vendored.** startergo's tree is unmaintained since 2026-01-14 and QEMU's GL scanout API changed to the "texture borrower" form (`DisplayGLTextureBorrower`). The 29-file mega-patch is trimmed to the 18 files this build actually compiles and forward-ported; dropped hunks are for GTK, SPICE, D-Bus and CoreAudio, none of which this runtime builds.
  - This also fixes a real runtime bug: `ui/cocoa.m` registered `&dcl_ops` rather than `&dcl`, so `dcl.ops = &dcl_gl_ops` was discarded and the display failed with *"Display cocoa is incompatible with the GL context"*.
- **Pinned Python build wheels** (`setuptools`, `wheel`, `pip`). QEMU 11.1 builds `qemu.qmp`, which needs `setuptools`; Homebrew's 3.14 has `tomllib` but no `setuptools`, and the CLT 3.9 has `setuptools` but no `tomllib`. Pinning sidesteps whichever interpreter is found.
- **Cocoa keyboard capture follows the key window** rather than the mouse grab, reconciled with the immersive-mode patch added in #50/#49 (see below).
- The audio patches from #59 and #60, which this branch is stacked on.
- A guest package-lock refresh, without which none of this can be built today.
- **The GPU-resolution patch is vendored too**, retiring the startergo source archive entirely (see below).

## Interaction with immersive mode

`qemu-cocoa-immersive-mode.patch` landed while this work was in progress and rewrites the same two Command-chord-forwarding sites. Both patches apply to 11.1.1 cleanly on their own; together they conflicted.

They are complementary rather than competing — immersive adds the *mode toggle*, this adds a fix for capture being tied to the mouse grab — so they are merged rather than one replacing the other:

```c
- (BOOL) isKeyboardCaptured
{
    if (!full_grab_enabled) {
        return NO;
    }
    return isMouseGrabbed || (eventsTap != NULL && [[self window] isKeyWindow]);
}
```

Standard fullscreen never captures, so the mode toggle behaves exactly as #50 intends. Immersive captures while the window is key rather than while the mouse happens to be grabbed: with an absolute pointing device, `notifyMouseModeChange` drops the grab as soon as the guest virtio-tablet driver binds, and `mouseExited` drops it again whenever the pointer leaves the view, handing host Command chords back to macOS mid-session.

`kVK_Command` and `kVK_RightCommand` move to the same predicate. Leaving them on `isMouseGrabbed` while the chords use the key-window predicate would toggle Super in the guest on a different condition than the chords that need it, stranding a held Super when the pointer leaves the view.

`run-qemu-gpu.sh` keeps the upstream `full-grab=$cocoa_full_grab` toggle rather than this fork's earlier hardcoded `full-grab=on`, so immersive mode stays user-selectable.

## Also here

- **`dtc` returns to `git.kernel.org`.** It had been temporarily mirrored while kernel.org's cgit was 404ing; it is serving again and the snapshot for the pinned commit hashes to the same `e115f987...`. Net effect is zero divergence from this repo's existing pin.
- **Guest package lock refreshed.** Eight packages had drifted (gstreamer + plugins, gvfs, libcdio, libcdio-paranoia, libvpx, mpv), so `make guest` refused the transaction on `main` as well. Six are pkgrel rebuilds, two are minor bumps, all in the media stack. Happy to split this out if you would rather review it separately, but without it the branch cannot be built.

## The startergo source archive is retired

`gpu-spike-resolution-fix.patch` was the last file the build still took from the `homebrew-qemu-virgl-kosmickrisp` tarball, so that archive is no longer downloaded, checksum-verified, root-validated or extracted at all. Two commits, deliberately separate:

1. **Vendor it byte-identical** (only a provenance header naming the source commit is added).
2. **Fix a bug in it.** The patch moved the `gl_dirty` clear out of `cocoa_gl_refresh` into `cocoa_gl_render`, but indented it inside the `if (cocoa_gl_trace_enabled())` block:

```c
    if (cocoa_gl_trace_enabled()) {
        glFlush();
        cocoa_gl_log_error("render end");
    /* Clear dirty flag after render completes to ensure clean handoff */
    gl_dirty = false;
    }
```

GL tracing is off in a normal build, so `gl_dirty` was never cleared. It latched true on the first damage and stayed true, and `cocoa_gl_refresh` re-rendered the scanout on every refresh tick for the rest of the session — a full texture blit per tick regardless of whether anything had changed. The code it replaced in `cocoa_gl_refresh` *did* clear the flag, so this came in with the patch rather than from QEMU. The indentation of the added lines gives the intent away: they sit one level out from the block they landed in.

Measured idle QEMU CPU fell from 26% of a core to ~15% (three samples: 15.0, 16.1, 15.3). Caveat on attribution: the before-run was an established desktop session and the after-run a freshly booted one, so some of that gap is the emptier desktop rather than the fix. The bug is wrong independently of what it cost.

Note this retires the *source archive* only. The virglrenderer, ANGLE and libepoxy release **binaries** still come from `startergo/homebrew-*`; those are separate repositories with separate pins and are untouched here.

## Verified

- The full nine-patch stack (now all nine in-repo bar none downloaded) applies to the pinned 11.1.1 tree (`c3d48b7d`, checksum-verified) with **zero rejects**.
- `make runtime`, `make guest` and `make app` all succeed; bundle is signed, `valid on disk`, and 17 Mach-O images verify for macOS 15.0.
- The VM boots to the Omarchy desktop, GPU acceleration active (`ANGLE Metal Renderer`), audio working.
- `./guest/test`, `macos-compatibility.test.sh`, `runtime-relocation.test.sh` pass.

## Not verified

Please weigh these in review:

- **Command-chord capture after the pointer leaves the view** — the exact case the keyboard-capture change targets. The test guest was a fresh image without the Hyprland binding needed to observe it, so this is verified by construction and code reading, not by a live keypress.
- The immersive/standard fullscreen toggle was not exercised end to end on 11.1.1.
- Window resize across a HiDPI boundary, switching the Mac output device mid-session, shared folder, clipboard, and a long-run RSS leak check.

The startergo tree being unmaintained is a standing risk this makes explicit: the vendored patch is now this repo's to maintain, which is the tradeoff for the display API having moved on.


themartiano:
Superb work.

The CI failure is from the old tests, on merge it should be fine.
Waiting to merge #60 before moving this one forward.

Fail-Safe:
I ran a controlled A/B on an M3 Max MacBook Pro (16 cores, 48 GB, macOS 27.0) and can corroborate the CPU improvement from this PR.

Tested PR head `c384be54`, locally integrated with main `04e5bf4` (`ab2e13f`). The guest kernel, initramfs, disk snapshot, 8-vCPU/4-GiB topology, Cocoa/ANGLE/virgl foreground display, and native audio/clipboard/camera bridge binary were held constant. The comparison was QEMU 10.2.50 + GICv2 versus QEMU 11.1.1 + GICv3.

| Metric (n=5 paired) | QEMU 10/GICv2 | QEMU 11/GICv3 | Change |
|---|---:|---:|---:|
| Settled QEMU CPU (% of one core) | 64.35% ± 0.90 | 11.90% ± 0.92 | **-52.45 pp / -81.51%** |
| Full runtime-stack CPU | 66.52% ± 0.87 | 14.90% ± 0.99 | **-51.62 pp / -77.60%** |
| Guest time to running Hyprland | 5.564 s ± 0.036 | 4.952 s ± 0.027 | **-0.612 s / -11.00%** |
| QEMU CPU consumed through boot | 13.939 CPU-s ± 0.223 | 3.989 CPU-s ± 0.053 | **-9.950 CPU-s / -71.38%** |

The paired 95% CI for the idle QEMU reduction was 51.06–53.84 percentage points. All five paired differences favored QEMU 11 (51.306, 52.990, 51.285, 52.822, and 53.840 pp), and the old/new ranges did not overlap (62.826–65.170% versus 11.330–13.534%). QEMU RSS was essentially flat/slightly higher (+29.4 MiB, +0.68%), so I would not claim a memory improvement.

Method: counterbalanced `ABBAABBAAB` order; fresh APFS clone of the same cleanly powered-off guest disk for every trial; guest readiness marker emitted after `Hyprland` existed; 30 s settling followed by a 60 s integrated CPU window. CPU time came from `proc_pid_rusage` with Mach timebase conversion, calibrated against a saturated core (2.015 CPU-s over 2 s wall while `ps` reported 100.0%). `100%` here means one CPU core.

This is still one host and five pairs, but the magnitude and consistency are substantive. It corroborates a large reduction in settled CPU, a modest wall-clock boot improvement, and a much larger reduction in CPU work during boot.

themartiano:
This PR is very relevant and new version has to be released soon, so I fixed the conflicts and I'm merging. 
Thank you!

caius72:
Thanks for merging this PR

PR #65, opened 2026-09-01
Apply backports to bin commands promoted to /usr/bin

BaruchEric:
## What broke

Fresh `make build` runs on current main fail during the guest build:

```
apply-omarchy-backports: target is not a regular file: bin/omarchy-install-service-1password
```

#58 added the first backport that targets a `bin/` path. `materialize-omarchy.sh` installs every Omarchy `bin/` command at `/usr/bin` inside the staged root and leaves package-path symlinks in `usr/share/omarchy/bin`, and backports apply after materialization (guest/tests/verify.py asserts that ordering). So the target is always a symlink by the time `apply-omarchy-backports.py` sees it, `contained_file` rejects it, and `git apply` could not patch through it anyway. The spec's `beforeSha256` matches the pristine upstream file at the pinned commit, confirming the patch is meant for the real file content.

## The fix

- `apply-omarchy-backports.py`: when a target inside `usr/share/omarchy` is a materialization symlink to `/usr/bin/<name>`, swap it for the real staged file, run the existing digest verification and `git apply` unchanged, then write the patched content back to `/usr/bin/<name>` and restore the symlink. The materialized layout comes out identical; targets that are regular files (like the `default/` and `shell/` ones) take the existing path.
- `write-provenance.py` had the matching problem one step later: its per-target `afterSha256` check read through the restored symlink, whose absolute `/usr/bin` link only resolves in the booted guest, so it raised FileNotFoundError. It now rebases absolute symlinks into the staged root before hashing. `digest_path` already hashed symlinks by link text and is unchanged, so tree digests are unaffected.

## Verified

- `pytest guest/tests/test_apply_omarchy_backports.py guest/tests/test_write_provenance.py` passes.
- A full `make build` on an M3 Mac gets past the guest authenticity steps with both the `1password-arm64-installer` and `notification-screen-privacy` backports applied and hash-verified, where it previously died at each of the two points above.

https://claude.ai/code/session_016yj5T57GoMLSn6dZCErGat

### DIFF
--- a/guest/scripts/apply-omarchy-backports.py
+++ b/guest/scripts/apply-omarchy-backports.py
@@ -6,6 +6,8 @@ from __future__ import annotations
 import argparse
 import hashlib
 import json
+import os
+import shutil
 import subprocess
 from pathlib import Path, PurePosixPath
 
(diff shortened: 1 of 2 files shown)


themartiano:
Thanks for the PR!

Codex found three issues:
1. stage_promoted_targets() can mutate before proving containment. A symlinked parent can cause it to unlink/copy outside omarchy_root before the later containment check rejects the target. Inputs are pinned, so this is not an immediate exploit, but it violates an important root-build invariant.
2. Restoration only happens on success. A bad preimage, failed patch, or postimage mismatch leaves the package-path symlink replaced by a regular file. Normal builds discard the failed rootfs, but --keep-rootfs and direct use retain an inconsistent tree.
3. The PR adds no tests. Existing tests only use regular files, so green CI does not exercise the promoted /usr/bin symlink path being fixed.

It also recommends:

The best near-term approach is to keep this design but make it transactional:
- Accept only exact bin/<name> -> /usr/bin/<name> mappings.
- Validate containment before mutation.
- Apply and verify in temporary regular copies.
- Commit verified bytes only on success and always restore symlinks.
- Add happy-path, provenance, failure-rollback, and containment tests.

A broader copy → patch → promote pipeline would be cleaner, but isn’t necessary for this fix.

BaruchEric:
Superseded by #70, which lands the transactional version of this fix (containment checked before mutation, patches applied to temporary copies, symlinks always restored) with tests. Thanks for the review and the quick follow-up; closing this one.

PR #80, opened 2026-09-01
feat: add signed Vivaldi ARM64 installer

Fail-Safe:
## Summary

- add Vivaldi to Omarchy browser install, remove, default, theme, and managed-policy workflows
- install Vivaldi Stable on ARM64 from one exact official RPM with pinned digest and signing identity
- repackage the authenticated payload as a deterministic Pacman-owned package while keeping Vivaldi out of the factory image

## Trust boundary

- user-initiated post-build install only; the proprietary payload is not redistributed with Try Omarchy
- exact HTTPS URL, version, RPM digest, official key digest, and signing fingerprint are pinned
- RPM signature, header and payload digests, package metadata, archive paths, required files, and installed binary identity are checked
- the generated package includes a Pacman `.MTREE`, enabling full installed-file verification
- Vivaldi documents Linux distribution integration at https://vivaldi.com/partners/linux/

## Validation

- `guest/test --source <clean pinned Omarchy v4.0.2 checkout>`: 54 tests and all guest/backport contracts pass
- disposable ARM64 Arch integration install as UID 1000: `vivaldi 8.2.4133.33-2`; 1,420 files, 0 altered according to `pacman -Qkk`, with root-owned installed paths
- native guest, QEMU runtime, and macOS app build completed with development signing
- strict app code-signature and macOS 15 compatibility checks pass
- QEMU port-forwarding, SSH contract, and power-action shell suites pass

## Local environment notes

- the production Swift target builds, but the local Swift test target cannot compile because the active Apple Command Line Tools installation lacks the `TestingMacros` plugin; non-Swift suites were run directly
- the persistent-storage suite reaches its final staged-QEMU assertion, which still hard-codes 10.2.50 even though current source and the staged runtime are both 11.1.1
- the current Arch Linux ARM mirror exposes `libbluray 1.5.1-1` while upstream still locks `1.5.0-1`; build validation temporarily matched the repository version, and this PR deliberately leaves that unrelated lock unchanged


themartiano:

This installs `vivaldi` with `pacman -U` but never adds it to a configured sync repository, so `pacman -Qem` treats it as foreign and Omarchy sends it through `yay -Sua`. That can route ARM64 guests to the x86_64-only AUR recipe instead of this pinned signed-RPM path. Please keep the package represented in a local sync repository, or explicitly exclude it from AUR updates, and cover the update path in the contract tests.

Fail-Safe:
Thanks — addressed in the latest push.

Vivaldi is now published into `/usr/share/try-omarchy/repo`, registered with `repo-add`, synced into pacman's `try-omarchy` DB, and installed with `pacman -S try-omarchy/vivaldi` instead of bare `pacman -U`. The installer also refuses to early-exit while `vivaldi` still appears in `pacman -Qem`, and asserts it is not foreign after install.

Contract tests cover the local-repo / non-foreign / `yay -Sua` path.

Issue #81, opened 2026-09-02
Feature request: configurable guest memory in the Start Menu


rdtiv:
# Feature request: configurable guest memory in the Start Menu

## Motivation

The guest always boots with 4 GiB of RAM (`-m 4G`, hardcoded in
`macos/run-qemu-gpu.sh`). On Macs with 16–64 GiB, users who run browsers and
dev tooling inside Omarchy push the guest into zram swap while the host has
plenty of free memory. vCPUs already adapt to the host (up to 8), but memory
does not.

## Current behavior

- `macos/run-qemu-gpu.sh` passes `-m 4G` unconditionally; the startup log
  states "4 GiB RAM" as a fixed fact.
- The guest runtime contract declares `minimumMemoryMiB: 2048` and
  `recommendedMemoryMiB: 4096`.
- The guest ships a zram swap device sized to RAM, which softens but does not
  remove the pressure.
- Nothing in the Start Menu or environment lets the user change the
  allocation; the balloon device cannot grow past the boot-time `-m` value.

## Proposal

Add a **Memory** row to the start menu, following the exact pattern the app
already uses for its other settings (UserDefaults preference store → an
`OMARCHY_QEMU_GPU_*` environment variable → re-validation in
`run-qemu-gpu.sh`). Since every launch begins at the start menu, one row
covers both moments that matter:

- **Configure at setup — the desired default behavior.** The row is present
  on the very first start menu, before the first boot, so users decide how
  much RAM Omarchy gets as part of initial setup. Doing nothing keeps
  today's exact behavior (4 GiB).
- **Adjust after setup — explicitly supported.** Because the start menu
  precedes every launch, changing memory later is the same one-click
  gesture: pick a new value, launch. This is technically trivial — memory is
  only a QEMU boot flag (`-m`), never baked into the guest image or the
  persistent VM data — so no reset, reinstall, or migration is ever
  involved.

The row:

```
Memory                                        4 GiB (default)
  Give Omarchy more of your Mac's RAM.
  Applies the next time Omarchy launches.
  ( 4 GiB · default )  ( 6 GiB )  ( 8 GiB )  ( 12 GiB )
```

- The row offers a small fixed set of choices — e.g. {4, 6, 8, 12, 16} GiB —
  filtered so only values that leave the host at least 8 GiB of headroom are
  shown (`hw.memsize` via `sysctl`). On an 8 GiB Mac the row is effectively
  read-only at 4 GiB.
[...]

PR #82, opened 2026-09-02
Add a configurable guest memory choice to the start menu

rdtiv:
Implements #81.

## What changed

The guest always booted with a hardcoded 4 GiB (`-m 4G` in
`macos/run-qemu-gpu.sh`). This adds a **Memory** row to the start menu so the
allocation is decided at setup and adjustable before any later launch, using
the same preference pattern as the other rows:

- `MemoryPreferences.swift`: `MemoryPolicy` (choices, host headroom rule,
  resolution), a `MemoryPreferenceStore` mirroring
  `FullscreenPreferenceStore`, and a `MemoryLaunchConfiguration` that exports
  `OMARCHY_QEMU_GPU_MEMORY_MIB`.
- Start menu: a Memory row (memorychip symbol + popup) between VM Location
  and Port forwarding. Choices are 4 (default) / 6 / 8 / 12 / 16 GiB,
  filtered so any non-default choice leaves the host at least 8 GiB; on an
  8 GiB Mac the row renders informational with the popup disabled. Each popup
  item carries its MiB value in its tag. A change applies on the next launch.
- `run-qemu-gpu.sh` re-validates the environment value independently: a
  whole number of MiB (1–7 digits, normalized to base 10 so a leading zero
  is never read as octal and a 20-digit value never wraps through 64-bit
  arithmetic), at least the guest manifest's 2048 MiB minimum, and — only
  above the default — small enough to leave the host 4 GiB. The value feeds
  `-m` and the two startup log lines.
- Reset launches strip the key like the other integration settings.
- The root README documents the row and the environment variable, including
  that the menu's 8 GiB headroom rule is deliberately more conservative than
  the launcher's 4 GiB safety floor; `macos/README.md` documents the
  UserDefaults → environment contract alongside the other settings.

## What deliberately did not change

- The default stays 4 GiB and still boots unconditionally, so existing
  installs and hosts under 8 GiB (including the 7 GB CI runners) behave
  exactly as before.
- No live resize: virtio-balloon cannot grow past the boot-time `-m`, and
  virtio-mem hot-plug is not worth the complexity here.
- A stored choice that no longer fits its host silently falls back to the
  default at launch, without rewriting the preference — move back to a
  bigger Mac and the choice returns.

## Tests

- `MemoryPreferencesTests.swift`: policy filtering/resolution, store
  round-trip and schema safety, launch configuration, and start menu
  presentation, in the existing Swift Testing style.
- `qemu-memory-contract.test.sh`: drives the real launcher with a fake QEMU
  and a shimmed `sysctl` — default `-m 4096M`, explicit 6144, fractional-GiB
  display, malformed value, below-minimum value, an oversized request on an
  8 GiB host, the default on a 7 GiB host (the CI-runner case, which fails
  if the host cap is ever applied to the default), a leading-zero value
  (octal-vs-decimal), and a past-64-bit value (wraparound).
- A cross-language contract test pins the environment key, default, and
  2048 MiB minimum between `MemoryPolicy` and the shell source.
- `QEMUGPULauncherTests` reset-sanitizer contract extended with the new key.
- Registered in the `make test` target.

## Review status

Reviewed and written from inside a running Try Omarchy guest (the ARM64
Arch VM this app boots — which is also why the feature exists: 4 GiB is
tight in here), against #81 and the surrounding launcher/start-menu/CI
paths. The guest has no Darwin toolchain, so `make test` was not run on a
Mac; GitHub `macos-15` will be the first Darwin execution of the new
Swift tests and `qemu-memory-contract.test.sh`. The shell validation was
exercised directly (bash octal/wraparound behavior confirmed by running
it), and both scripts pass `bash -n`.

The start menu, including Launch, is inside the existing scroll view, so
a short window should scroll rather than clip. Still worth a glance on a
13" to confirm Launch is visible without scrolling after the extra row
(760 → 832).

Headroom is installed RAM (`physicalMemory` / `hw.memsize`), not free
RAM, and does not count VirGL/Metal overhead on top of `-m`. That matches
#81. On a 16 GiB Mac the menu tops out at 8 GiB; the 12 GiB choice needs a
20 GiB host, so in practice it first appears on 24 GiB Macs.

## AI assistance

Developed with AI assistance, directed and reviewed by me:

- Research, implementation, and tests were drafted with Claude (Fable 5)
  via Claude Code, running inside the Try Omarchy guest.
- The diff was independently reviewed by Grok 4.6 and by a separate Claude
  multi-agent code-review pass; findings from both were verified before
  being applied or rejected — the octal/wraparound input hardening in
  `run-qemu-gpu.sh` came out of that review.
- I reviewed the full diff and take responsibility for the contribution.
  The commit carries a Co-Authored-By trailer.

## Reviewer notes

- `hostHeadroomMiB` (8 GiB for offering a choice) and the script's 4 GiB
  hard floor are judgment calls; happy to tune either.
- `qemu-memory-contract.test.sh` deliberately mirrors the self-contained
  harness of `run-qemu-ssh-contract.test.sh` rather than extracting a shared
  library, matching how each existing `*.test.sh` stands alone; happy to
  factor a common harness out if you'd prefer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ test:
 	@cd $(ROOT)/macos && SWIFT_MODULECACHE_PATH=$(ROOT)/macos/.build/module-cache/swift CLANG_MODULE_CACHE_PATH=$(ROOT)/macos/.build/module-cache/clang swift test --disable-sandbox
 	@$(ROOT)/macos/Tests/qemu-port-forwarding.test.sh
 	@$(ROOT)/macos/Tests/run-qemu-ssh-contract.test.sh
+	@$(ROOT)/macos/Tests/qemu-memory-contract.test.sh
 	@$(ROOT)/macos/Tests/qemu-power-actions.test.sh
 	@$(ROOT)/macos/Tests/qemu-persistent-storage.test.sh
 	@PYTHONDONTWRITEBYTECODE=1 python3 "$(ROOT)/macos/Tests/resize-vm-disk.test.py"
--- a/README.md
+++ b/README.md
@@ -291,6 +291,25 @@ or changing the enrolled Touch ID fingerprint set requires re-pairing. Disabling
 removes the guest enrollment and, while the host bridge is available, its wrapped
 Secure Enclave key representation.
 
+## Giving Omarchy more memory
+
+Use **Memory** on the start menu to pick how much of the Mac's RAM the guest
+boots with. The default is 4 GiB, and the menu only offers larger allocations
+(6, 8, 12, or 16 GiB) that leave macOS at least 8 GiB for itself, so an 8 GiB
+Mac shows the default alone. The choice is not tied to installation: change it
+before any launch, and it applies the next time Omarchy starts. Memory is a
+boot-time QEMU setting, never part of the guest image or VM data, so switching
+allocations never needs a reset and never touches your files. A stored choice
+that no longer fits the Mac it runs on falls back to the default.
+
+Scripted launches can set `OMARCHY_QEMU_GPU_MEMORY_MIB` (a whole number of
+MiB) instead. The launcher's own rule is looser than the menu's: it refuses
+values below the guest's 2048 MiB minimum, and values above the 4096 default
+that would leave the host under 4 GiB. The default itself always boots, and
+an environment value the menu would not offer (say 12 GiB on a 16 GiB Mac)
+is still accepted — the menu is deliberately conservative, the launcher is a
+safety floor.
+
 ## Requirements
 
 - Apple Silicon Mac (`arm64`)
--- a/macos/README.md
+++ b/macos/README.md
@@ -87,6 +87,16 @@ variable still wins, so the development and test override keeps working
 unchanged. Reset composes its environment exactly as a launch does, so it
 always erases the workspace the user is actually running.
 
+Guest memory follows the same preference pattern: the start menu's **Memory**
+row stores its choice in `UserDefaults` and publishes it to the launcher as
+`OMARCHY_QEMU_GPU_MEMORY_MIB`. The app only ever exports a value it resolved
+against this host (non-default choices must leave macOS 8 GiB), while the
+launcher independently enforces the guest's 2048 MiB manifest minimum and, for
+values above the 4096 default, a 4 GiB host floor — so a hand-set environment
+value gets the loose safety rule, not the menu's conservative one. Storage
+resets strip the variable like the other integration settings; memory is a
+boot-time `-m` allocation and never part of the guest image or VM data.
+
 Port forwarding is one versioned generic mapping list. The editor's **Add SSH**
 action only inserts the ordinary TCP `2222 → 22` preset; users may edit it like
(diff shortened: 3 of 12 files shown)


jamesyork-prog:
This looks like a solid foundation for configurable VM resources. I have a tested fork implementation that also exposes CPU count and initial disk size, but I do not want to compete with or duplicate this PR.

If the maintainers like the broader direction, I would be happy to build focused follow-ups after this lands:

1. CPU count as another per-launch preference, using the same host-aware policy/store/environment contract.
2. Initial disk size as a separate first-provisioning setting, retaining Try Omarchy's current 24 GiB default.

I would keep storage separate from runtime CPU/RAM because CPU and memory can be resolved on every launch, while disk size becomes coupled to the provisioned VM. That should also keep the memory review focused and make each follow-up easier to test and revert.

themartiano:
> This looks like a solid foundation for configurable VM resources. I have a tested fork implementation that also exposes CPU count and initial disk size, but I do not want to compete with or duplicate this PR.
> 
> If the maintainers like the broader direction, I would be happy to build focused follow-ups after this lands:
> 
> 1. CPU count as another per-launch preference, using the same host-aware policy/store/environment contract.
> 2. Initial disk size as a separate first-provisioning setting, retaining Try Omarchy's current 24 GiB default.
> 
> I would keep storage separate from runtime CPU/RAM because CPU and memory can be resolved on every launch, while disk size becomes coupled to the provisioned VM. That should also keep the memory review focused and make each follow-up easier to test and revert.

Yes, this is in my todo, users should be able to configure CPU and disk as well. CPU is `auto` by default capped at 8 cores if I recall correctly. Disk is fixed at 24gb. Ideally I'd say the disk should expand indefinitely, as needed. With a cap to not completely fill the mac storage.
What do you think?

rdtiv:
yes, i agree the comprehensive approach is ideal. 

Extending memory as proposed is a more focused step in the same direction, and unblocks memory limitations first to enable more exhaustive use. If you pursue a low risk targeted fix, I'd move first with memory. 

The comprehensive solution is ideal, the question is pace and surface area which I do not know your preference. Thank you so much for enabling Try Omarchy for us all.

themartiano (changes requested):
Hey,
Codex comments

The latest CI failure is reproducible against main: the new `qemu-memory-contract.test.sh` storage mock lacks `qemu_persistent_storage_select_existing` and `QEMU_PERSISTENT_STORAGE_MISSING_STATUS`, so QEMU never starts and `qemu.log` is absent. Update the mock for the current paired-boot contract (or reuse the current launcher harness), rebase onto #135, and rerun `make test`. The maintainer wants this PR before #156.

rdtiv:
@themartiano - i saw most recent comment, do you want me to take the actions or is someone else dong so? apologies in advance if your notes were intended for someone else. thank so much, -dan

rdtiv:
Rebased onto main (past #135) and updated the memory contract test's storage mock for the current paired-boot contract (`select_existing`, `MISSING_STATUS`, the nested-virtualization probe). `make test` passes locally in full, including the Swift suite (228 tests in 52 suites); CI should confirm once the workflow is approved.

PR #85, opened 2026-09-02
Add a VM backup before incompatible guest updates

patatman:
## Why

A new Try Omarchy guest build still cannot boot an existing disk. Reset is required, and the v0.2.0 notes already warn that `v0.1.0` disks are not migrated. Copying `~/Library/Application Support/Try Omarchy/VM/v1` in Finder is the wrong workaround: the working disk is a sparse ~24 GB file that Finder can inflate, and the 0600 workspace marker lands as 0644, which the launcher then refuses.

## What

- **Backup…** on the start menu copies recognized persistent disks into an empty APFS folder with `cp -c`, keeping sparse allocation and the 0700/0600 modes the launcher requires.
- The destination is a valid workspace for the **same** guest identity (choose it later as VM Location). It is not a migration onto a newer factory image.
- `make backup DEST=/path/to/empty/folder` does the same copy without rebuilding the app.
- Reset / incompatible-guest alerts tell you to back up first.
- README documents the two real update paths: export files (shared folder or SSH) for a new guest, or keep the whole VM for rollback to the matching app version.

Factory images under `images/` are not copied; the matching build rematerializes them.

## Tests

- `macos/Tests/qemu-persistent-storage.test.sh`
- Swift suites for launch flags, backup destination policy, and the start menu
- `macos/Tests/run-qemu-ssh-contract.test.sh`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ PACKAGE_NOTARY_PROFILE ?= $(RELEASE_NOTARY_PROFILE)
 FORCE ?= 0
 
 .DEFAULT_GOAL := help
-.PHONY: help doctor test guest runtime app build run run-ephemeral reset update-omarchy package package-preflight release release-preflight clean clean-all clean-guest
+.PHONY: help doctor test guest runtime app build run run-ephemeral reset backup update-omarchy package package-preflight release release-preflight clean clean-all clean-guest
 
 help:
 	@printf '%s\n' \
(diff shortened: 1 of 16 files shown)


patatman:
This is an automated PR created by an agent, verification is needed!

themartiano:
Thanks for the contribution!

Codex:
This branch predates #86. `main` now keeps each VM with its validated `boot/<identity>/` files and no longer requires Reset for a normal app update, while this backup copies only `disks/`. 

Me:
I'm closing this PR, since `v0.3.0` handles updates. If you think it could be better or want to add something, please feel free to open another PR. Thank you

PR #94, opened 2026-09-02
Keep development app builds out of Spotlight

jamesyork-prog:
## Summary

- move the generated app bundle under `dist/app.noindex/`
- remove and unregister the legacy `dist/Try Omarchy.app` on the first rebuild
- update build validation, launcher paths, and documentation for the new output layout
- cover the behavior in the existing focused build-cache test

## Why

A local build produces a complete app bundle. Spotlight can index it alongside an installed application, leaving Command-Space with two indistinguishable Try Omarchy results.

The earlier `.metadata_never_index` approach was not reliable for a `dist` directory that Spotlight had already indexed. A `.noindex` directory is macOS's dependable per-directory exclusion and still allows the bundle to run and be packaged normally.

The one-time legacy cleanup prevents an already-indexed bundle at the former path from remaining searchable after users pull this change.

## Verification

- `make test`
- `PYTHONDONTWRITEBYTECODE=1 python3 tests/test-build-cache.py`
- `bash -n macos/build-app.sh macos/open-qemu-gpu.sh`
- `git diff --check`
- built the pinned upstream QEMU 11.1.1 runtime and passed relocation and macOS 15 compatibility checks
- assembled the app under `dist/app.noindex/`, verified the legacy path was absent, and confirmed `mdfind` returned no generated-app result
- opened an ad-hoc-signed disposable app inside a `.noindex` directory and verified it registered with Launch Services while remaining absent from `mdfind`


themartiano:
Thank for the PR!

PR #95, opened 2026-09-02
Give the start menu a standard Mac app lifecycle

jamesyork-prog:
## Summary

- present the start menu with the regular macOS application activation policy
- add native About, Hide, Quit, Close Window, and Minimize menu commands
- return to accessory mode after QEMU reports that Omarchy is ready
- document the prelaunch behavior and cover its activation/menu contracts

## Why

The start menu is an interactive Mac window, but the helper currently runs as an accessory application from process launch. The red close control works, while standard application commands such as Command-Q, Command-W, and Command-M are not discoverable through a normal app menu.

This keeps the native lifecycle only for setup. Once the VM is ready, Try Omarchy returns to accessory mode so the Mac app chrome does not compete with the guest experience.

## Tests

- `make test`
- 183 Swift tests passed, including the new serialized application-presentation suite

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ port. The `dtc` mirror should be reverted once kernel.org returns.
 2. Open the DMG and drag **Try Omarchy** to **Applications**.
 3. Launch **Try Omarchy** from Applications.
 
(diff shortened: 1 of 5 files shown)


themartiano:
Nice, thanks

PR #96, opened 2026-09-02
Require typed confirmation before resetting the VM

jamesyork-prog:
## Summary

- require the exact text `Try Omarchy` before enabling the destructive Reset action
- present confirmation as a cancellable native sheet tied to the start window
- keep the existing unreachable-custom-folder recovery path unchanged because it does not erase a VM
- add coverage for casing and whitespace mismatches

## Why

Reset permanently removes the virtual machine, including its apps, files, accounts, and settings. A typed confirmation adds useful friction and makes accidental activation much less likely.

## Tests

- `make test`
- 182 Swift tests across 42 suites passed
- all shell and Python contract tests passed


themartiano:
Cool, thank you.
Did you accidentally erase yours? 😂

jamesyork-prog:
Luckily, no! But I’ve built enough software to know somebody was going to eventually 😂

PR #101, opened 2026-09-03
Pause the VM across macOS host sleep

### LINKED ISSUE #61: macOS host sleep terminates the VM instead of pausing it
**Summary**

When the macOS host goes to sleep, the VM is terminated outright rather than paused. The guest is not shut down and receives no notice — the QEMU process simply stops existing, and all in-guest state (running processes, terminal sessions, unsaved work) is lost. Reopening the app cold-boots the guest.

**Impact**

Any long-running work in the guest is lost on every host sleep. Closing the lid is effectively `kill -9` on the VM. In practice this means the app can't be left running between work sessions, which is at odds with the persistent-VM design (`~/Library/Application Support/Try Omarchy/VM/v1`).

**Evidence: the guest never sees it coming**

The guest journal ends mid-stream with no shutdown sequence at all. Across two separate occurrences, `journalctl -b -1` / `-b -2` contain **zero** lines matching `systemd-shutdown`, `Reached target Power-Off/Reboot/Shutdown`, or `Unmounting`:

```
IDX BOOT ID                          FIRST ENTRY                 LAST ENTRY
 -2 02aad6e945424eb99c3ee14293e1cbb6 Sat 2026-08-29 02:07:00 JST Sun 2026-08-30 07:42:16 JST
 -1 7fdc9db085fb41dc8a0a3c148287c5aa Sun 2026-08-30 08:55:02 JST Sun 2026-08-30 16:46:21 JST
  0 05c86c5ba2f940d9a70309efb26ba378 Sun 2026-08-30 19:13:24 JST ...
```

Both terminated boots end in the middle of an unrelated service's log output, with no final entry of any kind. There is no kernel panic, no oops, and no 9p/virtio error — the guest is simply frozen and discarded.

**Steps to reproduce**

1. Launch Try Omarchy and let the guest boot.
2. Start something observable in the guest (e.g. `tmux` with a long-running process).
3. Put the host to sleep (close the lid, or Apple menu → Sleep).
4. Wake the host.

Observed: the Try Omarchy app has quit. Reopening cold-boots the guest; the tmux session is gone.
Expected: the VM is paused before the host suspends and resumed on wake, with guest state intact.

**Suggested fix**

Register the launcher for `NSWorkspace.shared.notificationCenter` `willSleepNotification` and pause the VM before the host suspends, resuming on `didWakeNotification`:

```swift
NSWorkspace.shared.notificationCenter.addObserver(
    forName: NSWorkspace.willSleepNotification, object: nil, queue: .main
) { _ in /* pause vCPUs / QEMU `stop` */ }

NSWorkspace.shared.notificationCenter.addObserver(
    forName: NSWorkspace.didWakeNotification, object: nil, queue: .main
) { _ in /* resume / QEMU `cont` */ }
```

If wiring that into the Hypervisor.framework path is involved, even exposing a QEMU monitor/QMP socket would let users script `stop`/`cont` around sleep themselves. As far as I can tell the README documents no pause, suspend, snapshot, or monitor socket, so there is currently no workaround short of keeping the host awake with `caffeinate` or shutting the guest down before every sleep.

**Note on scope**

This is distinct from the bare-metal MacBook suspend issues discussed in basecamp/omarchy (e.g. #1840, #4184), which concern `/sys/power/mem_sleep` S3-vs-s2idle on T2 hardware. That is the *guest* suspending on real hardware. This report is about the *host* sleeping and taking the QEMU process with it on Apple Silicon — the guest's own suspend configuration is not involved.

**Environment**

- Try Omarchy: 0.4.0 (as reported by the app; the repo's releases page currently lists v0.2.0 as latest, so flagging in case the app's version scheme differs from the release tags)
- macOS: recent release on Apple Silicon (exact version available on request)
- Host: Apple Silicon MacBook Air (M1)
- Guest: Arch Linux ARM, kernel 7.2.0-2-aarch64-ARCH, aarch64, 8 vCPU, 3.8 GiB RAM
- Guest cmdline: `omarchy.qemu_virgl=1`

themartiano:
## Summary

- pause the VM synchronously over a private QMP connection when macOS announces host sleep
- resume only pauses owned by the host-sleep transition, with bounded reconnect retries after wake
- remove Cocoa Pause and Resume menu actions so pause ownership cannot become ambiguous
- fail closed when safe host-sleep control cannot be established

## QEMU compatibility

- based on current main after #62
- rebuilt the complete patched QEMU 11.1.1 runtime with the current Cocoa patch chain

## Testing

- make runtime
- make test
- 198 Swift tests across 45 suites
- guest and native contract tests
- staged-QEMU crash-lock test

Closes #61

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -36,6 +36,16 @@ is updated. Graphics travel from Linux through virtio-gpu and VirGL to the
 native Cocoa window. Storage, networking, audio, and input use their matching
 QEMU virtual devices and host backends.
 
+The macOS helper opens an authenticated connection to QEMU's private,
+single-client machine protocol socket before host sleep and retains that control
+session through wake. Before macOS sleeps it synchronously pauses the guest
+vCPUs, and after wake it resumes them only when that sleep handler observed the
+pause transition. The bundled Cocoa runtime removes its in-process Pause and
+Resume menu actions because they cannot participate in QMP connection
+ownership. Abnormal QEMU states such as an I/O error are never overridden. This
+preserves in-memory guest state across lid close while leaving safety stops
+untouched.
+
 One small host-integration channel sits beside those devices. A virtio-serial
 port (`dev.tryomarchy.clipboard`) carries newline-delimited JSON between a
 Swift bridge on the Mac, which watches the `NSPasteboard` change count, and a
--- a/macos/Sources/OmarchyVMHelper/FocusedCommandSuperBridge.swift
+++ b/macos/Sources/OmarchyVMHelper/FocusedCommandSuperBridge.swift
@@ -214,223 +214,39 @@ private extension CGEventFlags {
 }
 
 final class QMPMetaKeyClient: @unchecked Sendable {
-    private let descriptor: Int32
-    private let queue = DispatchQueue(label: "org.omarchy.qmp-meta-keys", qos: .userInteractive)
-    private var nextIdentifier: UInt64 = 1
-    private var closed = false
+    private let connection: QMPConnection
 
     init(socketPath: String) throws {
-        descriptor = try Self.connectSecureSocket(path: socketPath)
-        do {
-            guard let greeting = try Self.readJSONObject(from: descriptor, timeoutMilliseconds: 2_000),
-                  greeting["QMP"] != nil else {
-                throw HelperError.io("QMP socket did not send a valid greeting")
-            }
-            try Self.writeJSON([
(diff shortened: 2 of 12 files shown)


themartiano:

If a wake reconnect schedules a retry and the Mac sleeps again before it fires, the pending retry is not cancelled and can send `cont` during the next sleep. Please cancel pending wake retries in the sleep handler while preserving pause ownership, and add a re-sleep regression test.

PR #105, opened 2026-09-03
Fix GTK icon cache generation

Fail-Safe:
## Summary

- normalize copied Omarchy application artwork to the lowercase, hyphenated icon identifiers used by the corresponding desktop entries
- fail the guest build if two upstream artwork names normalize to the same target
- materialize the exact pinned Omarchy source in regression coverage and verify the resulting hicolor names and GTK cache-safety contract

## Problem

Omarchy 4.0.2 includes display-style artwork names such as `Google Messages.png`, `Disk Usage.png`, and `Retro Gaming.png`. Try Omarchy copied those basenames verbatim into `/usr/share/icons/hicolor/256x256/apps`, while the desktop entries reference `google-messages`, `disk-usage`, and `retro-gaming`.

Besides preventing those icon lookups from matching, the embedded spaces make `gtk-update-icon-cache` reject the generated cache. This surfaces during `omarchy-update` as:

```text
gtk-update-icon-cache: The generated cache was invalid.
error: command failed to execute correctly
```

The materializer now performs the same ASCII lowercase-and-hyphen normalization already reflected in the desktop icon identifiers.

## Validation

- Live v0.3.0 guest: reproduced the invalid cache with the six whitespace-bearing names; after applying the normalized mappings, `gtk-update-icon-cache -f -t /usr/share/icons/hicolor` reported `Cache file created successfully`, with no remaining cache-unsafe paths
- `guest/test --source <clean pinned Omarchy 4.0.2 checkout>`: 64 tests plus all guest/materialization contracts pass
- `make test`: build-cache tests, guest tests, macOS compatibility and relocation checks, 181 Swift tests across 42 suites, and QEMU networking/power/storage suites pass
- `git diff --check`, shell syntax, and Python compilation pass

`make build` was also attempted, but the guest build stops before materialization because the current Arch Linux ARM repository transaction has advanced beyond `packages.lock.json`. This PR deliberately does not mix an unrelated 24-package lock refresh into the icon fix.

### DIFF
--- a/guest/scripts/materialize-omarchy.sh
+++ b/guest/scripts/materialize-omarchy.sh
@@ -185,9 +185,15 @@ install_file 0644 "$source_dir/etc/fastfetch/config.jsonc" "$root/etc/fastfetch/
 # Preserve the application metadata and artwork used by Quickshell's real app
 # provider. Normalize display-style artwork names to the lowercase, hyphenated
 # icon identifiers used by the desktop files and accepted by GTK's icon cache.
+# The stem is normalized separately so Battle.net.png becomes battle-net.png
+# rather than battle.net.png.
 mkdir -p "$root/usr/share/icons/hicolor/256x256/apps"
 while IFS= read -r icon; do
-  icon_name=$(basename "$icon" | LC_ALL=C tr '[:upper:] ' '[:lower:]-')
+  icon_base=$(basename "$icon")
+  icon_stem=${icon_base%.*}
+  icon_ext=${icon_base##*.}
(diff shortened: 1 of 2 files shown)


themartiano:

`Battle.net.png` currently becomes `battle.net.png`, but its desktop entry uses `Icon=battle-net`, so this icon still will not resolve. Please map it to `battle-net.png` and cover that case in the regression test.

Fail-Safe:
Thanks — fixed.

The normalizer now rewrites the **stem** only (`Battle.net` → `battle-net`) and preserves the extension, so `Battle.net.png` materializes as `battle-net.png` to match `Icon=battle-net`. The regression contract requires `battle-net.png` and rejects both `Battle.net.png` and `battle.net.png`.

Also rebased onto current `main`.

PR #106, opened 2026-09-03
Guard omitted fcitx5 and bt-agent user units

### LINKED ISSUE #91: Guest image omits fcitx5 but still enables omarchy-fcitx5.service — infinite 203/EXEC loop floods 80% of the journal
### Summary

The guest image ships and enables `omarchy-fcitx5.service` but does not
install `fcitx5`. The unit has `Restart=always` / `RestartSec=2` and no
guard for a missing binary, so it respawns every 2 seconds forever.

```
NRestarts=14270
journal lines this boot:  89181
...matching fcitx5:       71356   (80%)
journalctl --disk-usage:  75M
```

Counts are from the moment I masked the unit, ~9h into a single boot.
Unmasked it grows at roughly 1,600 restarts/hour.

```
omarchy-fcitx5.service: Unable to locate executable '/usr/bin/fcitx5': No such file or directory
omarchy-fcitx5.service: Failed at step EXEC spawning /usr/bin/fcitx5: No such file or directory
omarchy-fcitx5.service: Main process exited, code=exited, status=203/EXEC
omarchy-fcitx5.service: Scheduled restart job, restart counter is at 14270.
```

The harm is diagnostic rather than functional — the guest works fine, but
the journal is unusable. I hit this while investigating an unrelated
Chromium crash, and the events around the crash were buried under tens of
thousands of fcitx5 lines.

### Why it happens

- `try-omarchy-runtime 4.0.1-1` owns
  `/usr/share/omarchy/default/systemd/user/omarchy-fcitx5.service`
  (pinned upstream `13f18b2cb7286fb54f87daf571a031aa6af3d8f0`).
- `fcitx5` is **available** in the guest's configured repos
  (`extra/fcitx5 5.1.21-1`) but is not installed in the image. It is one
  of 92 entries from `install/omarchy-base.packages` that are absent; I
  don't know whether that omission is intentional.
- Upstream `install/user/first-run/enable-user-units.sh` then runs
  `systemctl --user enable --now ... omarchy-fcitx5.service`
  unconditionally, and migration `migrations/1785167800.sh` re-enables it.
- The unit's only condition is `ConditionEnvironment=WAYLAND_DISPLAY`,
  which is satisfied. Nothing checks for the binary.

A stock Omarchy install has `fcitx5` in its base packages, so this does
not reproduce there — it is specific to the guest image.

### Suggested fix

A guest-side drop-in is the smallest change and needs no upstream patch:

```ini
# /usr/lib/systemd/user/omarchy-fcitx5.service.d/10-guard.conf
[Unit]
ConditionPathExists=/usr/bin/fcitx5
```

This matches an idiom upstream already uses in
`omarchy-tailscale-receive.service` (`ConditionPathExists=/usr/bin/tailscale`),
so it may be worth upstreaming too. Installing `fcitx5` in the image or
masking the unit would also work; the drop-in seemed least invasive and
keeps XCompose working if the package is added later.

### Related latent case in the same image

`bt-agent.service` is enabled by the same first-run script and its binary
(`/usr/bin/bt-agent`, from `bluez-tools`, also available in repos but not
installed) is absent too. It sits at `NRestarts=0` only because
`ConditionPathIsDirectory=/sys/class/bluetooth` is false in the VM — an
unrelated hardware guard, not a binary guard. If Bluetooth passthrough is
ever added, it will loop the same way.

Upstream basecamp/omarchy#6992 reports that same `bt-agent` loop from a
different trigger, which suggests auditing every unit in
`enable-user-units.sh` for a binary guard rather than fixing these singly.

### Environment

```
try-omarchy-runtime  4.0.1-1   (guest build 2026-08-25)
omarchy version      4.0.1-1
guest                aarch64, virtio-gpu, QEMU on Apple Silicon
host                 macOS (Apple Silicon MacBook)
```

Diagnosed on my machine with Claude Code (Claude Opus 5); reported by me.

Fail-Safe:
## Summary

- add `ConditionPathExists` drop-ins for `omarchy-fcitx5.service` and `bt-agent.service` so units enabled by Omarchy's first-run script stay inert when their binaries are absent from the trimmed ARM64 guest
- cover both guards in the guest contract suite

Fixes https://github.com/omacom/try-omarchy/issues/91

## Problem

The guest materializes and enables `omarchy-fcitx5.service`, but `fcitx5` is not in the factory package set. With `Restart=always` / `RestartSec=2`, the unit 203/EXEC-loops and can dominate the journal. `bt-agent.service` has the same binary gap; it only stays quiet today because `/sys/class/bluetooth` is missing in the VM.

## Approach

Match the idiom already used by upstream `omarchy-tailscale-receive.service`: guest-side drop-ins under `etc/systemd/user/*.service.d/` with `ConditionPathExists` on the ExecStart binary. Installing either package later makes the existing enabled unit start normally.

## Test plan

- [x] `python3 -m unittest discover -s guest/tests -p 'test_*.py'`
- [x] `python3 guest/tests/verify.py`
- [ ] After guest rebuild: `systemctl --user status omarchy-fcitx5.service` shows skipped/condition rather than restarting; journal no longer floods with 203/EXEC
- [ ] Optional: install `fcitx5` and confirm the unit becomes startable without changing the drop-in

### DIFF
--- a/guest/native-overlay/etc/systemd/user/bt-agent.service.d/10-guard.conf
+++ b/guest/native-overlay/etc/systemd/user/bt-agent.service.d/10-guard.conf
@@ -1,0 +1,5 @@
+[Unit]
+# bluez-tools is omitted from the trimmed guest. The unit already skips when
+# /sys/class/bluetooth is absent; guard the binary too so Bluetooth
+# passthrough cannot start a Restart=on-failure loop without bt-agent.
+ConditionPathExists=/usr/bin/bt-agent
--- a/guest/native-overlay/etc/systemd/user/omarchy-fcitx5.service.d/10-guard.conf
+++ b/guest/native-overlay/etc/systemd/user/omarchy-fcitx5.service.d/10-guard.conf
@@ -1,0 +1,5 @@
+[Unit]
+# The trimmed ARM64 guest omits fcitx5. Upstream enables this unit on first
+# run with Restart=always, so a missing binary would 203/EXEC-loop forever and
+# flood the journal. Skip cleanly until the package is installed.
+ConditionPathExists=/usr/bin/fcitx5
--- a/guest/tests/verify.py
+++ b/guest/tests/verify.py
@@ -427,6 +427,20 @@ def main() -> None:
         and "omarchy-provision-autologin-once.service" in native_autologin,
         "native provisioning keeps direct graphical login across VM boots",
     )
+    fcitx_guard = read(
+        GUEST / "native-overlay/etc/systemd/user/omarchy-fcitx5.service.d/10-guard.conf"
+    )
+    check(
+        "ConditionPathExists=/usr/bin/fcitx5" in fcitx_guard,
+        "fcitx5 user unit is inert until the omitted binary is installed",
+    )
+    bt_agent_guard = read(
+        GUEST / "native-overlay/etc/systemd/user/bt-agent.service.d/10-guard.conf"
+    )
+    check(
+        "ConditionPathExists=/usr/bin/bt-agent" in bt_agent_guard,
+        "bt-agent user unit is inert until the omitted binary is installed",
+    )
     check("omarchy-native-audio-bridge" in configure, "guest installs native host-audio integration")
     check(
         "default.target.wants/omarchy-native-camera-bridge.service" in configure,


themartiano:
Thank you 🚀

PR #138, opened 2026-09-05
[fix] Clarify aarch64 Install failures instead of target not found

Fail-Safe:
## Summary
- Keep unavailable Install-menu apps visible on Apple Silicon, but fail with a shared message (`The <app> app is not available via pacman/AUR for aarch64 at this time.`) instead of bare `target not found`.
- Central denylist + authenticity hooks in `omarchy-pkg-add` / `omarchy-pkg-aur-add`, plus Dropbox / GeForce NOW installer guards that stop before broken post-setup.
- Covered today: Ghostty, Edge, Spotify, Dropbox, Cursor, Grok Bot, LM Studio, Steam, Minecraft, Heroic, GeForce NOW; Battle.net/Lutris via installer-only refuses (not a global wine denylist).

Addresses #137 by clarifying the installation failure; this does not make Ghostty available on ARM64.

Related to #57 and #18: this improves the Install failure UX for known x86-only apps, but it is not a full fix for every Super+Space / package failure (repo/mirror issues, already-ported apps like 1Password/Vivaldi, etc.).

### Follow-ups
- Trim the denylist once [omarchy-pkgs#240](https://github.com/omacom/omarchy-pkgs/pull/240) ships an aarch64 `[omarchy]` channel (several entries here—Ghostty, Cursor, Heroic, LM Studio, Grok Bot—are expected to become installable then).
- Separate tiny PR: add **Install → Browser → Vivaldi** menu row (install path from #80 exists; menu entry does not).
- Optional later: Y/n “hide this menu item” and/or ⚠ glyph overlays.

## Test plan
- [x] Ephemeral second VM (SSH 2223; reserved Codex VM untouched)
- [x] Ghostty / Minecraft / Cursor / Spotify / Heroic / Edge / Steam / Battle.net / GeForce NOW / LM Studio show the shared message
- [x] Dropbox stops at the message (no tray setup / `dropbox-cli` not found)
- [ ] CI green on this PR

Issue #137, opened 2026-09-04
Installing Ghostty doesn't work


mohamedmansour:
Says `error: target not found` I am on Apple M4 Pro.

plainprince:
same error happens on installing Ollama, I'm on M2 Max

Fail-Safe:
`target not found` for Ghostty (and similarly Ollama) on Apple Silicon is the aarch64 Install gap: those packages are not available via pacman/AUR for this guest today, so the Install path fails with a bare pacman error.

**PRs:**
- https://github.com/omacom/try-omarchy/pull/138 — keep the menu entries, but fail with a clear shared message instead of `target not found`
- https://github.com/omacom/try-omarchy/pull/140 — optional follow-up to hide unavailable Install rows after that message (Y/n)

Until those land / a new image ships them, Ghostty / Ollama Install from the menu will keep failing on aarch64; they are not a broken Omarchy update path.

themartiano:
#138 has now merged. It replaces the bare Ghostty installation failure with a clear message that the app is unavailable through the guest's aarch64 package path.

It does not add a working ARM64 Ghostty package or make installation succeed. Leaving this issue open for actual package support.

NimbleAINinja:
Until there's an aarch64 package, you can build Ghostty from source inside the guest. I'm running 1.3.1 this way on two Try Omarchy guests (Apple Silicon). The steps are the same as in #221.

**1. Get Zig 0.15.2.** Ghostty 1.3.x needs exactly 0.15.x. The repos ship 0.16 and the AUR `zig0.15` package is x86_64-only, so use the upstream binary:

```sh
mkdir -p ~/ghostty-build && cd ~/ghostty-build
curl -LO https://ziglang.org/download/0.15.2/zig-aarch64-linux-0.15.2.tar.xz
tar xf zig-aarch64-linux-0.15.2.tar.xz
export PATH="$PWD/zig-aarch64-linux-0.15.2:$PATH"
```

**2. Install the system gtk4-layer-shell and build.** Don't use `-fno-sys=gtk4-layer-shell`. That bundles the library, and the binary ends up pointing into `.zig-cache`.

```sh
sudo pacman -S --needed gtk4-layer-shell
curl -LO https://release.files.ghostty.org/1.3.1/ghostty-1.3.1.tar.gz
tar xf ghostty-1.3.1.tar.gz && cd ghostty-1.3.1
[...]

Issue #57, opened 2026-08-29
"target not found" when installing any package


hoverlover:
I get this error every time I try to install any package.  I tried `omarchy update` and also rebooting but nothing helps.

<img width="1750" height="1200" alt="Image" src="https://github.com/user-attachments/assets/72237b45-2ae2-45e8-9962-3950e766e259" />

themartiano:
ARM package compatibility is being worked on for Omarchy in general, will be fixed

timharris777:
I ran into this today-- is there an official place where this is being worked on. I searched through the omarchy repo but couldn't find any clear direction on how they are looking to tackle this. Would love to contribute if possible.

themartiano:
Yes, this repo and PR: https://github.com/omacom/omarchy-pkgs/pull/240
It's not the only piece, but it's a big one.

themartiano:
Related progress: merged PR #138 replaces bare target-not-found errors with a clear aarch64 availability message for known unsupported Install-menu apps. It improves the explanation but does not supply the missing ARM packages or resolve every package/repository failure.

Leaving this issue open because the broader installation problem is not fully resolved by that PR.

Issue #18, opened 2026-08-26
Some super space “commsndd “ don't work?


ramarivera:
First of all, kudos, it'd amazing how really the whole process is download and click.
Second: is it possible there are some rough edges, mostly around package installation or super space “commands”?

E.g. 

Tried to install 1pass, asks for sudo, red error text flashes the floating window and it gets closed.
Installed tailscalr, logged in. However the tray icon still claims tailscale is missing?
Tried to install chrome, no luck.
Tried to change the default browser settings, changes are ignored.

It's totally valid that you tell me I missed some very obvious step but it's my first time with Omarchy

themartiano:
No, I think those are real issues, probably stemming from some common source(s). I'll investigate

ramarivera:
Lmk if I can share more info or logs or whatever,

ramarivera:
Hey hey @themartiano . 

I have some unexpected free time on my hands, so please let me know if any idea to debug this come to mind. I just need a bit of pointing in the right direction :)

ramarivera:
<img width="814" height="661" alt="Image" src="https://github.com/user-attachments/assets/c39fd85d-ea8a-4d8c-9bab-2b7fd5ffe879" />

Could u share when you plan to release a new version? Asking cuz it seems some of the very recent fixes could be the fix to my issue[s] as well 🤞🏻

themartiano:
Today.
I found a small annoying UI glitch that I'm fixing, then I'm releasing!
It's a much better exp

themartiano:
> Hey hey [@themartiano](https://github.com/themartiano) .
> 
> I have some unexpected free time on my hands, so please let me know if any idea to debug this come to mind. I just need a bit of pointing in the right direction :)

Sorry for not replying before, I've been on a rush since the release.
As you said, some of those issues should be handled on the new version. Can you please take a look once it's released?
Also, if you want to contribute and need help with setup etc send me a DM on x.com/martiano, easier to talk over there, unless it's something specific about a PR etc.
Thank you!

ramarivera:
Sure thing, just ping me the moment its ready and I'd be glad to help!
Re contributing, thats a good point, i will work to get the environment up and running and i'll reach out should something n ot work

themartiano:
@ramarivera it's out!
https://github.com/themartiano/try-omarchy/releases/tag/v0.2.0

ramarivera:
[...]

PR #140, opened 2026-09-05
[feat] Proposal: opt-in hide for aarch64-unavailable Install apps

Fail-Safe:
## Summary

@themartiano — proposed UX path for aarch64 Install failures on Try Omarchy (seeking your direction before treating this as merge-ready).

Today many Install-menu apps still fail with bare `target not found`. [#138](https://github.com/omacom/try-omarchy/pull/138) keeps those rows visible and fails with a shared message. This PR **includes that foundation** and proposes a small next step for decluttering:

1. Show the shared unavailable message (no fake `Done!`).
2. Opt-in prompts (default **N**):
   - `Hide <app> from the Install menu? [y/N]`
   - `Hide all known unavailable Install apps too? [y/N]`
3. Persist under `~/.local/state/try-omarchy/`.
4. Setup → **Restore hidden Install apps** to undo.
5. Expand the denylist for Install paths that still leaked `target not found` in local testing (RetroArch cores spam, Sublime, VSCode, Zed, NordVPN, ONCE, ChatGPT Desktop, Xbox Controllers, Bitwarden, Ollama, …).

Glyph overlays were tried locally and dropped (truncation / misaligned labels). Hide-on-refuse felt like the better product shape.

If you prefer the message-only change first, land/keep [#138](https://github.com/omacom/try-omarchy/pull/138) and we can rebase this on top — or close #138 in favor of this single proposal.

## Test plan
- [x] Ephemeral second VM (SSH 2223; reserved Codex VM untouched)
- [x] Steam / Minecraft / Ghostty: message + hide prompts; no `Done!`
- [x] Hide one / hide all → rows disappear from Install (+ Super+Space)
- [x] Setup → Restore hidden Install apps brings them back
- [x] RetroArch / Sublime / Ollama: shared message instead of `target not found`
- [ ] Maintainer UX sign-off on prompt copy + whether hide-all belongs in-flow vs Setup-only

Addresses #137; related to #57 and #18. Includes the message-only foundation from #138.

### DIFF
--- a/guest/native-overlay/usr/local/bin/omarchy-arch-aarch64
+++ b/guest/native-overlay/usr/local/bin/omarchy-arch-aarch64
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# omarchy:summary=Returns true when the guest architecture is aarch64
+# omarchy:hidden=true
+
+[[ $(uname -m) == aarch64 ]]
--- a/guest/native-overlay/usr/local/bin/omarchy-install-has-hidden
+++ b/guest/native-overlay/usr/local/bin/omarchy-install-has-hidden
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# omarchy:summary=True when the user has hidden any aarch64-unavailable Install items
+# omarchy:hidden=true
+
+/usr/local/bin/omarchy-arch-aarch64 || exit 1
+
+state_dir="${XDG_STATE_HOME:-$HOME/.local/state}/try-omarchy"
+[[ -f $state_dir/hide-all-aarch64-unavailable && ! -L $state_dir/hide-all-aarch64-unavailable ]] && exit 0
+[[ -f $state_dir/hidden-install-packages && ! -L $state_dir/hidden-install-packages ]] || exit 1
+grep -q '[^[:space:]]' "$state_dir/hidden-install-packages"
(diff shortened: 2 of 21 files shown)


Chessing234 (changes requested):
solid ux direction, but the wine-related denylist entries look too broad for a global pkg-add hook — they can block aarch64 installs that never hit the install menu. tighten to install-path-only or drop deps that actually exist on aarch64.

Fail-Safe:
@Chessing234 Agreed — that denylist scope was too wide.

In d42644b:
- Removed `umu-launcher` / `wine-staging` / `wine-mono` / `wine-gecko` from the global `aarch64-unavailable-packages` denylist (so pkg-add no longer blocks them elsewhere).
- Kept Install-menu hide ids for `battlenet` / `lutris`.
- Added installer-only early refuses on `omarchy-install-gaming-battlenet` and `omarchy-install-gaming-lutris` via the shared unavailable helper.

Happy to adjust further if you’d rather drop Wine refusal entirely and leave only the menu hide.

Chessing234 (approved):
wine/umu are off the global denylist and only refused from the battlenet/lutris installers — that was the concern. looks good.

themartiano (commented):
Hey, I think it's better to warn people about missing packages (#138) but I think we should keep them visible.
We want to fix missing packages soon, so I'd prefer to keep them visible as our goal is to have all of them working.
What do you think, we merge #138 and close this?

Fail-Safe:
@themartiano Agreed. Since the plan is to fix these soon, keeping the rows visible is the right default. Hiding them would just make the gap easier to ignore.

Let's merge #138 and I'll close this. Thanks!

Issue #133, opened 2026-09-04
Feature request: Touch ID / Apple Keychain support for fingerprint unlock


rohrig:
Hey, cool project and thank you for what you're doing. It would be great if the fingerprint reader worked for unlocking the Apple Keychain. Right now it asks me to insert a key instead of prompting for my fingerprint, and I'd love to see that supported.

Fail-Safe:
*Preface: I am not associated with this project in any official capacity, so my response is strictly my opinion and not meant to give any indication as to the project's actual roadmap.*

With that out of the way, as an enthusiast of this project, I love the idea and would love to see this work. However, this is a lot more challenging than it may appear at first.

My understanding is Touch ID isn't a USB fingerprint reader. It is cryptographically paired with the Mac's Secure Enclave. Applications receive an authentication result, not fingerprint data or a device that QEMU can forward to the guest OS. [Apple documents this boundary explicitly here.](https://support.apple.com/guide/security/biometric-security-sec067eb0c9e/web)

Not sure if you're familiar with the Asahi Linux project for Mac, but even with the incredible level of bare-metal/native functionality they were able to tap into for the M1 & M2 series hardware, Touch ID is still called out as an unusable function.

Who knows, though. Maybe there's a devil in the details somewhere that someone can find that changes this story. 🙂

rohrig:
I've got it working. Let me know if you want a pull request. As I say in the video, it probably not the best code so I don't want to submit it and make you review slop. But if you'd like it for inspriation, I'll gladly share it. https://youtu.be/7h7I0aQ4D-M

rohrig:
I should add, it does not use the Apple keychain directly. It does use the built-in fingerprint reader, but you have to create new passkeys as if it's a separate computer. So I think the comment above does like have some validity to it. But if you're just wanting to be able to log into site with the passkey, this is possible.

Fail-Safe:
@rohrig Awesome--kudos! I would love to see the code you've got for your working demo if you're able to share it. Do you have a fork with a branch you could link here?

themartiano:
Related progress: merged PR #143 adds opt-in Touch ID authentication for guest sudo, configured through Setup → Security → Touch ID for sudo.

Its scope is sudo authentication only. It does not expose Apple Passwords/Keychain, implement browser passkeys, or replace guest login/screen-unlock authentication. Leaving this issue open for the requested Keychain/passkey integration.

themartiano:
[...]

PR #143, opened 2026-09-06
[feat] Add opt-in Touch ID authentication for guest sudo

Fail-Safe:
## Summary

- add an opt-in Touch ID authentication bridge for guest `sudo`
- add a stateful **Setup → Security → Touch ID for sudo** control with enable, test, re-pair, and disable actions
- keep the integration dormant until the user authenticates with the guest sudo password and approves enrollment with Touch ID
- preserve normal guest-password authentication as fallback and display the reason before requesting the password
- enable guest time synchronization at boot so clock drift does not invalidate otherwise valid signed approvals

This implements the separate guest-sudo feature. It does not attempt to expose Apple Passwords/Keychain data or replace guest login and screen-unlock authentication.

## Security model

- the guest authentication device and enrollment state are root-only
- each approval uses a fresh request ID and 256-bit challenge
- the Secure Enclave P-256 signature binds the operation, guest identity, PAM user, requesting user, service, TTY, key identity, and 15-second validity window
- the guest pins the enrolled public key and verifies every signature with OpenSSL
- macOS prompts with `deviceOwnerAuthenticationWithBiometrics`, biometric-set binding, no password fallback, and zero reuse duration
- the host prompts only while the matching QEMU process is alive and focused
- enabling writes the PAM rule only after enrollment succeeds; disabling removes the PAM rule before deleting enrollment state
- existing recognized PAM rules migrate to enable user-visible fallback explanations without re-enrollment; approval validity checks remain unchanged

## Validation

- `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer make test` passed, including guest unit and contract tests, 210 Swift tests in 47 suites, macOS compatibility/runtime relocation, and QEMU forwarding, power, SSH, and persistent-storage tests
- the release host helper builds successfully; new regression coverage checks compact menu insertion, command failure propagation, refusal of password fallback, host prompt cancellation, late callbacks, and late-response recovery
- regression coverage verifies that a real signed approval is rejected with a clock explanation when guest time is outside the allowed window, while successful authentication stays quiet
- checked message delivery through the running guest's actual PAM library: the explanation reaches the conversation and failed authentication remains rejected
- live guest validation: time synchronization corrected approximately 16 seconds of drift; Mac and guest clocks then matched, and the user confirmed Touch ID sudo authentication without a guest password
- verified enable, disable, re-enable, and persistence across shutdown/power-on during the original implementation
- the original security review covered commit `c523029`; it is not presented as a security review of the subsequent time-sync and diagnostic changes
- clean factory-image rebuild attempted: guest package resolution stopped with `could not satisfy dependencies` before reaching the changed configuration scripts. For live testing, cached guest artifacts were refreshed and the app rebuilt; signature and macOS compatibility checks passed. This does not establish a successful clean factory build.

### Deliberate manual failure-path testing

Tests were performed interactively with real Mac Touch ID and the persistent guest. Each authentication test cleared sudo's cached credentials.

- **Normal approval:** with guest time synchronized, fingerprint approval completed `sudo true` with exit status 0 and no guest password.
- **Explicit cancellation:** canceling Touch ID displayed the fallback explanation; entering the correct guest password then completed ordinary sudo with exit status 0.
- **Focus loss:** switching to another Mac application while the prompt was open, then successfully scanning a finger, resulted in rejected approval and a guest-password prompt. The dialog remained visible and repeatedly returned to the foreground; automatic cancellation on focus loss is not claimed.
- **Timeout on `dc06bcf`:** two unanswered attempts both closed the Touch ID dialog and reported failed authentication without accepting password fallback. The first was timed at exactly 55 seconds; the second was not timed. No numeric exit status was captured for these timeout attempts.
- **Recovery after timeout on `dc06bcf`:** two subsequent interactive test runs completed without a guest password, each reporting exit status 0, with no mismatched-request errors.

Normal approval, explicit cancellation, and focus-loss checks were performed before the final timeout/test-command update. Timeout and recovery were repeated after installing the `dc06bcf` Mac helper and guest scripts in the enrolled VM. Earlier runs against the old installation reproduced the stale-response failure and are not counted as validation of the fix.

Remaining UX limitation: host timeout still appears as “Touch ID was not approved”; the shared broker message mentions password fallback even though the dedicated test deliberately refuses it.

## Operational boundaries

- supported scope is guest `sudo` only
- moving or resetting the VM, changing Macs or macOS accounts, or changing the enrolled biometric set may require re-pairing
- the Try Omarchy window must be focused when Touch ID is requested
- new factory images enable `systemd-timesyncd`; existing guests with drift can enable it with `sudo systemctl enable --now systemd-timesyncd.service` and verify synchronization with `timedatectl`
- unanswered Mac prompts are canceled after 55 seconds; the guest discards late responses without extending its 65-second deadline
- the test and enable/re-pair checks refuse password fallback and propagate failure; passwordless sudo policies are explicitly distinguished from evidence of Touch ID use

#133 remains separate: this PR does not implement Apple Keychain unlock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,24 +249,34 @@ Loopback binding prevents devices on Wi-Fi, Ethernet, or the wider LAN from
 connecting. It does not isolate the listener from other users or processes on
 the same Mac; guest SSH authentication is still required.
 
-### Experimental Touch ID for sudo
+### Touch ID for sudo
 
-The experimental native authentication bridge can enroll this Mac and use
-Touch ID as a sufficient authentication method for guest `sudo`:
+The native authentication bridge can enroll this Mac and use
+Touch ID as a sufficient authentication method for guest `sudo`. Open
+**Omarchy Menu → Setup → Security → Touch ID for sudo**, or run:
 
 ```sh
-try-omarchy-touch-id-enroll
-try-omarchy-touch-id-test
+try-omarchy-touch-id
 ```
 
-Enrollment requires the normal guest sudo path once, then pins a public key for
-a Touch ID-protected Secure Enclave signing key. The Mac stores only the
-Secure Enclave's device-bound encrypted key representation. Every later approval is signed
-over a root-private guest ID, fresh challenge, the sudo user and requesting user,
-the interactive TTY, and a 15-second validity window. Each enrolled guest has a
-distinct host signing key. The QEMU window must be frontmost. Cancellation,
-invalid responses, missing enrollment, and unavailable Touch ID all fall back
-to the normal guest password; no login or screen-unlock PAM policy is changed.
+The integration ships disabled. Enabling first requires the normal guest sudo
+password, then Touch ID creates and proves possession of a Secure Enclave
+signing key. Only after that succeeds is the narrowly scoped sudo PAM rule
+installed. The menu then offers Test, Re-pair, and Disable actions.
+
+The Mac stores only the Secure Enclave's device-bound encrypted key
+representation. Every later approval is signed over a root-private guest ID,
+fresh challenge, the sudo user and requesting user, the interactive TTY, and a
+15-second validity window. Each enrolled guest has a distinct host signing key.
+The QEMU window must be frontmost. Cancellation, invalid responses, missing
+enrollment, and unavailable Touch ID all fall back to the normal guest password;
+no login or screen-unlock PAM policy is changed.
+
+Enrollment persists across guest and Mac restarts for the same persistent VM,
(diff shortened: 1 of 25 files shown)


themartiano (approved):
Thanks for the pull request!

Codex comments

Reviewed enrollment, signed request binding, password fallback, and host-process checks. Full tests and pinned-source checks pass with the merged changes. Guest sudo is separate from #133, which remains open for Keychain support. Live checks are documented in the PR; a clean factory rebuild remains blocked by the existing package-resolution problem.

PR #150, opened 2026-09-07
Add runtime tablet/mouse toggle for relative pointer lock

### LINKED ISSUE #144: fix: Mouse Capture
Mouse capture / relative mouse input is broken

Mouse capture in Try Omarchy is currently broken. The cursor cannot be reliably captured by applications that request pointer locking or relative mouse input, causing it to escape the VM window instead of staying captured.

This makes applications requiring continuous mouse movement, especially games, difficult or impossible to use properly.

Expected: Applications should be able to capture the mouse and receive relative pointer movement normally.

Actual: Mouse capture/pointer locking fails, and the cursor just sits there.

Fail-Safe:
## Summary
- Keep absolute `virtio-tablet` as the desktop default, with stable QMP ids (`omarchy-tablet` / `omarchy-mouse`) so the pointer can hot-swap without restarting the VM.
- Add a Start Menu **Relative mouse** preference for launch, plus a menu-bar mouse control while the VM is running (the start menu dismisses at launch).
- Document that Cocoa `full-grab` is focused keyboard Super capture, not relative mouse lock — games need relative `virtio-mouse` before pointer lock works (fixes #144).

## Test plan
- [ ] Launch with Relative mouse off: desktop trackpad pointing stays absolute; confirm `virtio-tablet` / `id=omarchy-tablet` in the QEMU command line.
- [ ] From the menu-bar mouse icon, switch to **Game pointer**, then verify a pointer-lock / relative-mouse app or game can capture the cursor (Ctrl-Alt releases).
- [ ] Switch back to **Desktop pointer** and confirm absolute pointing returns without restarting.
- [ ] Enable **Relative mouse** on the start menu, relaunch, and confirm the session starts on relative mouse.
- [ ] `DEVELOPER_DIR=…Xcode… swift test --filter Pointer` and `macos/Tests/run-qemu-ssh-contract.test.sh`


Fail-Safe:
Closing — mid-session UI and QMP hotplug need a different approach before this is ready for review.

PR #153, opened 2026-09-07
Replace themartiano paths with omacom

eluhr:
The repo moved to the omacom org in v0.3.0, but some references still point at themartiano/try-omarchy. Old paths redirect, so nothing is broken, but they should name the current owner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ port. The `dtc` mirror should be reverted once kernel.org returns.
 
 ## Quick start
 
-1. Open [Releases](https://github.com/themartiano/try-omarchy/releases) and download the latest signed and notarized `.dmg`.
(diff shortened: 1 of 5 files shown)


themartiano (changes requested):
Nice, thanks for the PR.

Codex comments

Changing `supplyChain.hyprland.issue` changes the component digest, but `macos/run-qemu-gpu.sh` still pins the old digest. `guest/test` fails at “native launcher accepts and pins the patched Hyprland component,” and newly built images would be rejected. Recompute the launcher digest from the updated metadata and rerun `make test`.

eluhr:
@themartiano Repinned the dependency in aee0a7d and ran `make test`. The full test suite now passes, including the guest contract check that was previously failing.

themartiano (approved):

PR #141, opened 2026-09-05
feat: add field to initial setup to set the vm disk size

semeano:
Addresses #104 and #132 for new/reset VM capacity; existing-disk growth is tracked in #154.

Issue #132, opened 2026-09-04
More disk space


kensaso1959:
Could we please have more disk space? 30 GB isn't much; 100 GB or 250 GB would be perfect.

plainprince:
I'd do it just like #81 : have another row in the start menu titled "Disk Space"
For now though, use a shared folder.

themartiano:
Merged PR #154 provides a command-line way to grow a supported existing VM to a larger capacity, including 100 or 250 GiB, without resetting its contents. The maintenance command is macos/resize-vm-disk.sh in a source checkout; it supports a preview and an explicit --apply step that retains a verified backup. It requires a stopped VM, APFS, Python 3, sufficient host space, and a valid paired boot kit.

This does not yet provide the requested start-menu control. #141 remains open for new/reset VM capacity selection, and #104 has a report of the resize tool rejecting an older VM without the expected boot directory. Leaving this open while those gaps remain.

Issue #104, opened 2026-09-03
Feature request: Choose Omarchy disk size


valentin2105:
Hello Sir,

Thanks for this awesome project, waiting to native Omarchy on M3 mac.

It would be awesome to be able to give more than 25Gb to the Omarchy VM.

two more question : 

- When you ship a new release, how to update without destroying our omarchy setup ? 
- It seem I cannot upgrade the system : 

```


Update Arch signing keys
Keys are correct

Update system packages
:: Synchronizing package databases...
 core is up to date
 extra is up to date
 alarm is up to date
 aur is up to date
 omarchy is up to date
 try-omarchy is up to date
:: Starting full system upgrade...
 there is nothing to do
warning: linux-aarch64: ignoring package upgrade (7.2-2 => 7.2.2-2)

Update mise tools
mise All tools are up to date


Restarting shell
All plugins have been reloaded
```
 
Thanks you

martijnwip:
I'd also like to see this land — being able to set the VM disk size (or grow an existing one) would be very helpful.

  One data point that might be useful: I tried the merged resize-vm-disk.sh from #154 against my existing VM (created Aug 29,
  before that merge) and it fails immediately:

  state boot directory is not a direct directory: .../VM/v1/boot

  My VM predates the newer "boot kit" metadata layout that the resize script expects, and there doesn't seem to be a migration
  path yet to bring an older VM's on-disk state up to that layout. So right now, existing users on older installs can't use the
  new resize tool even after updating the app. Might be worth tracking that migration gap alongside #141 (which only covers
  new/reset VMs).

  Happy to test against my setup if a migration path gets added.

themartiano:
Partial progress: merged PR #154 adds macos/resize-vm-disk.sh to grow a stopped, supported existing VM from a source checkout while retaining a verified backup. It requires APFS, Python 3, and a valid paired boot kit; it does not add a disk-size picker to the app.

The older-VM boot-directory failure reported above still needs follow-up before this can be considered covered for that installation. #141, which adds capacity configuration for new/reset VMs, remains open. Leaving this issue open for the remaining capacity-selection and compatibility work.

PR #154, opened 2026-09-07
[feat] Add safe growth for existing VM disks

Fail-Safe:
## Summary

Existing VM disks default to 24 GiB, and choosing a larger factory-image capacity does not enlarge an existing installation. Add a source-checkout maintenance command that grows a stopped VM without resetting its data:

```sh
macos/resize-vm-disk.sh --size-gib 32          # preview
macos/resize-vm-disk.sh --size-gib 32 --apply  # back up and grow
```

The command reuses the launcher's workspace lock, metadata checks, and paired-boot validation. It requires APFS, checks host headroom, retains checksum-verified disk and boot backups, and extends only the validated disk inode. It refuses active VMs, shrinking, unsafe paths, and unrecognized storage. Equal-size requests are no-ops. The existing guest boot service expands ext4 on the next launch.

Custom locations use an explicit `--state-root`; this command does not read the app's location preference or support legacy development workspaces. It needs Python 3 but no app rebuild. The README covers capacity limits, backup retention, and recovery.

Related to #104 and #132. Complements draft #141, which configures capacity for new/reset VMs and explicitly excludes resizing existing disks. This provides a command-line maintenance path that a future UI can build on.

## Validation

- Full `make test` with the Xcode developer directory selected, including 14 new resize tests.
- `bash -n` and `git diff --check`.
- Ran the command on an isolated APFS clone of an existing 24 GiB guest: verified backup, growth to 32 GiB, successful boot with the original paired kernel, automatic ext4 growth, and clean shutdown.
- A second resize was rejected while real QEMU held the workspace lock.
- Read-only `e2fsck -fn` passed after shutdown; filesystem metadata reports 8,388,608 blocks of 4,096 bytes (32 GiB).

No factory-image inputs or native app code changed. The isolated boot used an existing built runtime; this change does not require a new factory build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ test:
 	@$(ROOT)/macos/Tests/run-qemu-ssh-contract.test.sh
 	@$(ROOT)/macos/Tests/qemu-power-actions.test.sh
 	@$(ROOT)/macos/Tests/qemu-persistent-storage.test.sh
+	@PYTHONDONTWRITEBYTECODE=1 python3 "$(ROOT)/macos/Tests/resize-vm-disk.test.py"
 
 guest:
 	@OMARCHY_FORCE_BUILD="$(FORCE)" "$(BUILD_CACHE)" \
--- a/README.md
+++ b/README.md
@@ -285,6 +285,47 @@ all of that app's factory-image changes to an existing VM, and an in-guest
 update should not be assumed to reproduce them. A confirmed reset is the
 deliberate, destructive way to start again from the newest bundled factory.
 
+### Growing an existing VM disk
+
+To add capacity without resetting the VM, shut down Omarchy and run the
+maintenance command from a source checkout on the Mac:
+
+```sh
+# Preview a new total capacity of 32 GiB.
+macos/resize-vm-disk.sh --size-gib 32
+
+# Retain a verified backup, then enlarge the stopped disk.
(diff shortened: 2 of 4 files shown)


themartiano (approved):
Thanks for the pull request!

Codex comments

Reviewed locking, backup verification, inode checks, and growth-only behavior. Full tests passed, including all 14 resize tests. Existing live APFS validation is documented.

martijnwip:
Tried this on an existing VM created before v0.3.0 (with the older disks/images/locks layout, no boot/ directory). After updating to v0.4.0 and launching, the app runs the one-time boot-file pairing and it fails every time with:

Try Omarchy could not complete the one-time boot-file pairing. The saved VM was not reset or upgraded. You can safely try again.

Running omarchy-vm-helper directly from Terminal shows the actual guest boot log:

qemu-persistent-storage: reusing saved VM 95ea56e2483f; the bundled factory image applies only to new VMs
[qemu-gpu] Pairing the saved VM with its original boot files (one time).
...
:: running late hook [try-omarchy-boot-export]
ERROR: Try Omarchy boot export source is invalid
run-qemu-gpu: the one-time boot recovery did not publish an exact, complete export

This VM previously showed an ignored kernel package upgrade (7.2-2 => 7.2.2-2) during a normal in-guest system update, might be related to why the boot export can't find a valid source. The disk otherwise boots fine on the older app version. Happy to share more logs if useful.

PR #135, opened 2026-09-04
Give the start menu Omarchy styling

jamesyork-prog:
## Summary

- apply the official Omarchy dark palette and terminal typography to the start menu
- organize permissions and integrations into themed cards
- replace Apple-styled launcher actions with keyboard-accessible Omarchy buttons
- present immersive mode as an accessible `[ ON ]` or `[ OFF ]` control
- keep Try Omarchy branding, existing behavior, and upstream attribution unchanged

## Verification

- `make test`
- rendered the complete launcher document in a real AppKit window and checked the themed cards, statuses, reset action, launch action, and immersive toggle

Addresses #97.

PR #156, opened 2026-09-07
Add CPU and memory settings to the VM start menu

nocell:
The VM currently always starts with up to 8 CPU cores and 4 GiB of RAM. Add a **Resources → Configure…** sheet so users can choose both values before launching, including all host cores and allocations such as 12 GiB, without editing or re-signing the app.

The sheet supports Save, Cancel, and Use Defaults. It persists both settings in versioned UserDefaults, displays the effective allocation in the start menu, and passes it to QEMU for persistent and ephemeral launches. Existing defaults are preserved. CPU choices range from 4 to the host count; RAM uses whole GiB starting at 2, with allocations above 4 GiB leaving 4 GiB for macOS. Saved values that no longer fit a smaller host fall back independently without overwriting the saved preference.

The shell launcher validates canonical decimal values and host limits before touching VM storage. App launches replace inherited resource overrides with the displayed selection; storage-only resets strip the resource settings. The start menu's preferred height grows with the new row and remains capped to the screen with scrolling.

Validation:
- Full `make test` passed under Python 3.12 (`uv run --no-project --python 3.12 make test`): 210 Swift tests plus guest and shell suites.
- Final `swift test --disable-sandbox` and native release build passed.
- Launcher tests cover defaults, 18 cores / 12 GiB, boundaries, smaller hosts, malformed/overflowing input, rejection before storage access, and ephemeral launches.
- An isolated native UI smoke check confirmed the saved 18-core / 12-GiB summary after reopening and those values at a recording launcher boundary. A complete new guest/runtime image was not built; packaging could not reuse the installed older guest artifact because its supply-chain schema differs from current main.

Related: #81, #82.


themartiano (commented):
Great, this is very useful.

We will first merge #82, which handles memory, then merge this PR.
In the meantime, we merged #135 that changes the styling of the start menu, so please rebase and update the new config UI to match the styling.
Thanks.

themartiano:
Rebased and updated the UI.
Merging now, thank you

PR #166, opened 2026-09-08
Persist the guest console instead of discarding it

### LINKED ISSUE #165: Guest console output is discarded: stdio chardev inherits the app's /dev/null stdout
## Summary

The guest console is wired to a `stdio` chardev, and QEMU inherits the helper app's stdout. For an app launched from Finder that stdout is `/dev/null`, so **every byte the guest prints to its console is discarded on every normal launch** — kernel boot messages, oopses, panics, and early-userspace failures alike. When a guest fails to boot, loses its network, or hangs, there is no record of it anywhere on the host, and whatever evidence existed is gone the moment the VM restarts.

## Evidence

The console device is created against a `stdio` chardev:

```
macos/run-qemu-gpu.sh:1424
  -chardev 'stdio,id=omarchy-hvc0,signal=off'
  -device 'virtconsole,bus=omarchy-serial.0,nr=0,chardev=omarchy-hvc0'
```

and `console=hvc0` is a required part of the kernel command line (`macos/run-qemu-gpu.sh:754`), so this really is where guest console output goes.

That chardev's stdout comes from the helper, which passes its own through:

```swift
// macos/Sources/OmarchyVMHelper/QEMUGPULauncher.swift:758-760
process.standardInput  = FileHandle.standardInput
process.standardOutput = FileHandle.standardOutput
process.standardError  = pipe
```

For a GUI launch that is `/dev/null`, which is observable on a running VM:

```console
$ lsof -p <qemu pid> | awk '$4 ~ /^[012][a-z]*$/ {print $4, $5, $9}'
0r CHR /dev/null
1u CHR /dev/null
2  PIPE
```

`stderr` is a pipe the helper reads, but the console is not on stderr — it is on stdout.

## Impact

I hit this while trying to diagnose a guest that was alive but had wedged networking: QEMU idling at the normal ~15% CPU, a valid DHCP lease, `info usernet` showing only stale flows and two sockets stuck in `CLOSE_WAIT`, and `systemd-resolved` falling back to its `1.1.1.1` fallback server. A reboot cleared it — and with it, any chance of finding out why, because the console had gone to `/dev/null` and the VM's own journal was the only remaining copy.

This is the class of fault where the console is the primary evidence, and it is exactly when it is unavailable.

## Suggested fix

QEMU chardevs already accept `logfile=`, and the bundled runtime supports it:

```console
$ 'Try Omarchy.app/Contents/Resources/runtime/bin/Try Omarchy' -chardev stdio,help
  logappend=<bool (on/off)>
  logfile=<str>
```

Pointing the existing chardev at a file under the VM's own state directory captures the output with no change to how the console is attached. Keeping one previous generation matters, since a fault serious enough to force a reboot is only readable *after* that reboot has already replaced the live log.

Verified end to end against the bundled runtime, booting the bundled kernel with no root disk so it prints and then fails:

```console
$ ls -l cons2.log
-rw-r--r--  1 user  wheel  92610 cons2.log
$ head -3 cons2.log
[    0.179835] usbcore: registered new interface driver uas
[    0.179925] usbcore: registered new interface driver usb-storage
[    0.180001] usbcore: registered new interface driver ums-alauda
```

92,610 bytes of guest console output captured where nothing is captured today.

PR follows.

## Environment

- Try Omarchy v0.3.0 (release DMG), macOS 26.6.2, Apple M4 Pro
- Reproduces on `main` @ 768d03d

jtavin-smp:
Fixes #165.

## Problem

The guest console is bound to a `stdio` chardev (`macos/run-qemu-gpu.sh:1424`), and the helper hands QEMU its own stdout (`QEMUGPULauncher.swift:759`). For an app launched from Finder that is `/dev/null`, so every console byte is discarded on every normal launch. `stderr` is piped and read by the helper, but the console is not on stderr.

Confirmed on a running VM:

```console
$ lsof -p <qemu pid> | awk '$4 ~ /^[012][a-z]*$/ {print $4, $5, $9}'
0r CHR /dev/null
1u CHR /dev/null
2  PIPE
```

The practical cost is that a guest which fails to boot, wedges its network, or hangs leaves no host-side record, and a restart destroys what was there.

## Change

Give the existing chardev an explicit `logfile=`, pointed at the VM's own state directory. Nothing about how the console is attached changes.

- **Persistent launches** log to `<VM root>/console.log`, next to the VM they describe, so the log survives app updates exactly as the rest of that directory does.
- **One generation is rotated aside** to `console.log.1` at launch. This is the part that makes the feature useful: a fault bad enough to force a reboot is only readable after that reboot has already truncated the live log.
- **Ephemeral launches** log into the temporary workspace and discard it with the rest of that state.
- The option value is comma-escaped, since QEMU splits chardev options on commas and a user-selected VM location may contain one.

`logappend=off` plus the rotation keeps this bounded at two files rather than growing without limit.

## Verification

End to end with the bundled runtime, booting the bundled kernel with no root disk:

```console
$ ls -l cons2.log
-rw-r--r--  1 user  wheel  92610 cons2.log
$ head -3 cons2.log
[    0.179835] usbcore: registered new interface driver uas
[    0.179925] usbcore: registered new interface driver usb-storage
[    0.180001] usbcore: registered new interface driver ums-alauda
```

92,610 bytes captured where nothing is captured today.

`macos/Tests/run-qemu-ssh-contract.test.sh` passes, with two added assertions: the chardev carries the logfile option, and a pre-existing `console.log` is rotated to `console.log.1` with its contents intact.

The test's fake storage shim previously set `QEMU_PERSISTENT_STORAGE_DIRECTORY` but not `QEMU_PERSISTENT_STORAGE_ROOT`, even though its own `-kernel "$persistent_root/boot/kernel"` assertion treats `FAKE_PERSISTENT_ROOT` as the root. This adds the missing assignment so the shim reflects what the real module exports.

## Note on `make test`

`make test` does not currently complete in my environment: `swift test` fails with `no such module 'Testing'` (48 occurrences). This is **not** caused by this change — it reproduces identically on a clean `main` worktree at 768d03d with no modifications, and this PR touches no Swift sources. Every shell and Python test in the suite that runs before it passes. Flagging it rather than claiming a green run.

## Docs

README gains a short "Guest console log" section under *Data and updates*, covering both paths, the rotation, and a note that the log contains guest output and is worth reviewing before attaching it to a bug report.


themartiano:
Merged main, fixed some CI issues and merging.
Thank you!

PR #168, opened 2026-09-08
Add native hardware video decoding, HDR output, and audio continuity

### LINKED ISSUE #167: Add host-backed hardware video decoding for the ARM64 guest
The README documents CPU-only video decoding. The ARM64 guest has accelerated VirGL graphics, but applications cannot use the Mac's media decoder through the current virtual devices.

Proposed implementation within the existing Apple Silicon / QEMU / HVF architecture:

- A guest VA-API driver and a broker restricted to the `video` group send compressed HEVC, AV1 and VP9 to a supervised host helper.
- VideoToolbox must confirm a hardware decoder before the driver exposes the session.
- A private Mach capability carries IOSurfaces to QEMU for GPU copies into the guest's existing VirGL textures. CPU readback uses bounded per-client buffers through a dedicated PCI aperture.
- Application-scoped, pinned FFmpeg libraries support mpv and Firefox without replacing Arch's FFmpeg libraries. Firefox keeps its RDD sandbox; Vivaldi provides the verified VP9 YouTube path.

A local M5 Pro installation completed the requested 2160p60 YouTube clip without playback stalls or dropped frames during a 7,529-frame measurement interval. HEVC Main10 also works in mpv and sandboxed Firefox. Initial FFmpeg validation matched all 120 decoded 4K Main10 frames and timestamps against software decoding; the accompanying PR will include reproducible checks and separate decode-only timings from CPU readback costs.

Opening this issue before the implementation PR as requested by `CONTRIBUTING.md`, so the new guest/host transport and application integration can be reviewed explicitly.

nocell:
ARM64 guest applications spend substantial CPU time decoding video, while the existing eight-bit display path loses HDR output. This adds host VideoToolbox decoding, a paired ten-bit HDR presentation path, consistent SDR interface brightness, and a fix for reproduced HDA audio dropouts.

Closes #167. CPU/RAM settings remain separate in #156.

## Demonstration

The two owner-supplied recordings show the same **Tron 4K60 HDR source**. The accelerated setup uses a **120 Hz guest display**; this is not a claim of 120 FPS source playback. The recordings are resized to 1920×1242 for upload, retain their original timing and audio, and preserve ten-bit BT.2020/PQ capture metadata.

**Before — hardware video decoding and guest HDR disabled** (42.64 seconds):

https://github.com/user-attachments/assets/c18d8b6b-b707-4ea9-8d85-a152888c7806

**After — hardware video decoding and native HDR enabled** (60.52 seconds):

https://github.com/user-attachments/assets/38d4f88c-223e-43b5-9b64-92a1ad33c71c

These are visual demonstrations, not an isolated decoder benchmark: both acceleration and HDR differ, and the window layouts differ. Both files use the Mac's HDR screen-recording format, including the recording of the SDR guest. Playback appearance depends on the viewer's browser and display. Neither recording measures physical panel luminance or proves frame delivery at 120 FPS.

## Implementation

- Hardware-required VideoToolbox sessions support HEVC Main/Main10, AV1 Main and VP9 profiles 0/2. A supervised VA-API broker/helper exchanges bounded messages; private Mach capabilities pass decoder IOSurfaces to QEMU/ANGLE for GPU copies into VirGL textures. A bounded memory path supports CPU readback. Firefox retains its RDD seccomp sandbox.
- The paired virtio-gpu module exposes ten-bit scanout and HDR metadata. Hyprland composites BT.2020/PQ; QEMU presents through an RGBA16Float Metal/EDR layer. Shared GPU events and a bounded three-surface pool avoid CPU readback and completion waits. Drawable acquisition runs off QEMU's main loop; an activity token prevents App Nap throttling while allowing normal system sleep.
- The guest enables HDR only after capability negotiation. Scoped Mesa/mpv builds handle P010 and Wayland color metadata; Vivaldi enables Wayland color management. Older guest drivers keep the SDR path.
- The active host display preset supplies an SDR-white hint, with a 250-nit fallback. A Hyprland patch advertises the same white level to color-managed clients, correcting dim Chromium interfaces beside brighter wallpaper. Chromium can also adapt HDR midtones to this white level; mpv's explicitly targeted PQ output remains unchanged in the tested comparison.
- SDL's audio timer changes from 10 ms to 1 ms, preventing the reproduced full HDA-ring drops. PipeWire's existing quantum stays at 4096.
- Sources, patches, binary hashes and corresponding-source notices are included in the native guest packages and build integration.

## Validation

Development machine: M5 Pro, 18 vCPUs and 12 GiB guest RAM. Measurements are specific to the tested streams and machine.

- Final `make test` passed with Python 3.12: 219 Swift tests in 48 suites, 87 guest unit tests, guest contracts, build/cache tests and macOS shell/storage suites. Native SDR-white and client-white policy tests passed separately.
- Rebuilt and verified the QEMU runtime, native-video/HDR packages and patched Hyprland. The actual Metal presenter test passed SDR → HDR → SDR transitions, 100/1000-nit signal values, primaries, orientation, GPU synchronization and bounded submission under a stalled presentation queue.
- FFmpeg: all 480 frames across four HEVC/AV1/VP9 streams matched software decoding exactly, including timestamps. GPU surface reuse and CPU readback were tested separately. Three alternating decode-only repetitions produced the medians below; guest CPU time excludes the host helper and is not total system power.

| Decode-only stream | Hardware FPS | Software FPS | Hardware guest CPU s | Software guest CPU s |
| --- | ---: | ---: | ---: | ---: |
| HEVC 4K, 10-bit | 56.68 | 43.89 | 0.794 | 14.318 |
| AV1 4K, 10-bit | 115.16 | 17.34 | 0.403 | 103.364 |

Hardware decoding reduces guest CPU work but is not always faster for every codec. The AV1 software result is below real-time 60 FPS; this benchmark uses a different fixture from the Tron recordings.

- Foreground playback: the complete 126.56-second 4K60 YouTube test had zero drops across 7,529 measured frames. mpv's complete 4K AV1 test had one presentation drop and zero decoder drops. Subsequent HDR checks covered YouTube VP9 Profile 2 4K60, local AV1 10-bit 4K60 and HEVC Main10 4K30, with media time checked against wall time.
- Digital scanout checks confirmed Chromium white moving from approximately 205 to 597 nits for a 600-nit host hint, HDR highlights near 1000 nits, and unchanged explicitly targeted mpv PQ pixels across the SDR-white change. These are signal checks, not colorimeter readings.
- Audio A/B during muted 4K60 YouTube playback: the 45-second stereo tone changed from 37 discontinuities/channel and 43.3787 seconds retained audio to zero discontinuities and 45.0000 seconds. QEMU and SDL callback captures agreed.
- A clean factory image was assembled from an unprovisioned verified base, current overlays and verified packages, finalized/repacked with project scripts, and booted on a new user disk through graphical setup. No previous user state was imported. Source changes and added commit history were checked for private workspace references and common credential patterns before publication.

## Remaining validation and tradeoffs

This remains a draft for architecture/implementation review. A full from-source Docker guest build is still blocked because pinned Rust `1:1.98.0-1` disappeared from the current Arch ARM repository; the clean-image validation reused the verified existing ttfx package without changing that pin. It does not establish a complete from-source rebuild.

HDR uses a private paired virtio extension and an exact `7.2.2-2-aarch64-ARCH` kernel module. Optional private CoreDisplay APIs supply only a display capability hint and fall back safely when unavailable. The HDR surface pool adds about 190 MiB at 4K, excluding Metal drawables. Dolby Vision dynamic metadata is not transported. Other Mac generations, external displays, Bluetooth and physical acoustic output need separate validation. The 1 ms audio timer increases requested wakeups while audio is active.

See [native video](https://github.com/nocell/try-omarchy/blob/38f7f2b7011002eb72d781147f9eb95960efba62/docs/native-video.md) and [native HDR](https://github.com/nocell/try-omarchy/blob/38f7f2b7011002eb72d781147f9eb95960efba62/docs/native-hdr.md) for contracts, reproduction commands and detailed results.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-# Blank context lines, including the final hunk's context, are patch syntax.
-*.patch whitespace=-blank-at-eol,-blank-at-eof
+# Blank lines and a space before context indentation are patch syntax.
+*.patch whitespace=-blank-at-eol,-blank-at-eof,-space-before-tab
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
-*.patch whitespace=-blank-at-eol
+# Blank context lines, including the final hunk's context, are patch syntax.
+*.patch whitespace=-blank-at-eol,-blank-at-eof
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Omarchy's trademark rights.
 
 - Hardware-accelerated ARM64 virtualization and VirGL graphics
 - Nested KVM virtualization on M3 and newer Apple Silicon
+- Hardware video decoding through the Mac's VideoToolbox media engine
 - Resizable native window with automatic guest resolution and HiDPI scale updates
 - Mac audio input/output selection inside Omarchy, with live routing and system-default fallback
 - FaceTime HD and other Mac cameras exposed to Omarchy as an on-demand 720p webcam
@@ -25,7 +26,10 @@ Omarchy's trademark rights.
 - One optional shared Mac folder, available inside Omarchy under the same name (`~/Work` stays `~/Work`)
 - Loopback-only TCP and UDP port forwarding from the Mac into Omarchy
 
-> **Current limitation:** Video decoding is CPU-only, so playback can be slow, especially at high resolutions. An improved video path is in development.
+The native video path supports HEVC, AV1 and VP9 in the bundled mpv integration,
+HEVC in Firefox, and VP9 YouTube playback in Vivaldi. Available codecs depend on
+the Mac's hardware. See [native video](docs/native-video.md) for application
+setup, verification and measured performance limits.
 
 ## Changes in this fork
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Omarchy's trademark rights.
 - Hardware-accelerated ARM64 virtualization and VirGL graphics
 - Nested KVM virtualization on M3 and newer Apple Silicon
 - Hardware video decoding through the Mac's VideoToolbox media engine
+- HDR10 output for YouTube in Vivaldi and local files in mpv
 - Resizable native window with automatic guest resolution and HiDPI scale updates
 - Mac audio input/output selection inside Omarchy, with live routing and system-default fallback
 - FaceTime HD and other Mac cameras exposed to Omarchy as an on-demand 720p webcam
@@ -30,6 +31,8 @@ The native video path supports HEVC, AV1 and VP9 in the bundled mpv integration,
 HEVC in Firefox, and VP9 YouTube playback in Vivaldi. Available codecs depend on
 the Mac's hardware. See [native video](docs/native-video.md) for application
 setup, verification and measured performance limits.
+See [native HDR](docs/native-hdr.md) for the paired display driver, HDR output
+verification and kernel compatibility requirements.
 
 ## Changes in this fork
 
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -12,7 +12,7 @@ code.
 - **Arch Linux ARM packages** — each package retains its own license. The
   generated package transaction is recorded in `packages.lock.txt`.
 - **Hyprland** — BSD-3-Clause; the reviewed v0.56.1 source and rounded-border
-  coverage backport are pinned in `guest/spec.json`. The guest package retains
+  coverage and client SDR-white patches are pinned in `guest/spec.json`. The guest package retains
   Hyprland's upstream license and dependency metadata.
 - **Glaze** — MIT; the pinned v7.2.0 headers are used by the Hyprland build, and
   their verified upstream license is retained in the rebuilt guest package.
@@ -24,6 +24,13 @@ code.
   the guest under `/usr/share/try-omarchy/native-video-source`; its GPL notice
   is in `/usr/share/licenses/try-omarchy-native-video`. The system FFmpeg
   package remains separate. Linked Arch packages retain their own notices.
+- **Linux virtio-gpu, Mesa 26.2.1 and mpv 0.41.0** — the native-HDR package
+  contains a patched GPL-2.0-only kernel module, Mesa under its component
+  licenses (primarily MIT), and mpv under GPL-2.0-or-later plus linked-library
+  obligations. Pinned source archives/files, patches and the build recipe are
+  included under `/usr/share/try-omarchy/native-hdr-source`; license notices
+  are in `/usr/share/licenses/try-omarchy-native-hdr`. Source identities are
+  recorded in `guest/hdr/sources.json`.
 - **PyYAML and packaging** — pinned build dependencies used to build
   VirGLRenderer and QEMU, respectively; MIT for PyYAML and Apache-2.0/BSD for
   packaging. They are not loaded by the application at runtime.
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -18,6 +18,15 @@ code.
   their verified upstream license is retained in the rebuilt guest package.
 - **ANGLE, VirGLRenderer, libepoxy, SDL, libslirp, GLib, Pixman, and other QEMU
   dependencies** — retain their respective upstream licenses.
+- **FFmpeg 9.0.1** — the private native-video build enables GPL and version 3
+  components and is distributed under GPL-3.0-or-later. Its pinned source
+  archive, local bitstream-preservation patch and build recipe are included in
+  the guest under `/usr/share/try-omarchy/native-video-source`; its GPL notice
+  is in `/usr/share/licenses/try-omarchy-native-video`. The system FFmpeg
+  package remains separate. Linked Arch packages retain their own notices.
+- **PyYAML and packaging** — pinned build dependencies used to build
+  VirGLRenderer and QEMU, respectively; MIT for PyYAML and Apache-2.0/BSD for
+  packaging. They are not loaded by the application at runtime.
 - **mise** — MIT; the reviewed ARM64 release is pinned in `guest/spec.json`.
 - **ttfx** — MIT; the reviewed source release and locked Rust dependencies are
   pinned in `guest/spec.json` and built natively for ARM64. Its packaged
--- a/docs/native-hdr.md
+++ b/docs/native-hdr.md
@@ -1,0 +1,156 @@
+# Native HDR output
+
+The native QEMU display can present HDR10/PQ video from Vivaldi's YouTube player
+and local files opened with `mpv`. This extends the hardware decoder described in
+[native video](native-video.md): preserving ten decoded bits alone does not make
+an eight-bit SDR desktop an HDR display.
+
+## Display path
+
+The paired guest driver exposes ten-bit RGB scanout and the DRM `Colorspace` and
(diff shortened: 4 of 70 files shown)


Fail-Safe:
@nocell && @themartiano:

Tested revision `38f7f2b7011002eb72d781147f9eb95960efba62` on an M3 Max with an isolated preview app and VM. There are some encouraging results here: the hardware decoder bridge works across the tested codecs, the HDR guest/presenter path initializes, and native Wayland presentation feedback reaches approximately 120 fps.

These results are from a local integration with build adjustments and a repaired guest dependency set, rather than a clean build of the unmodified PR.

## Build fixes

### 1. Require the DRM VA-API backend in the private FFmpeg build

Our initial FFmpeg build completed, but hardware decoding failed with `No VA display found`. The configure link check could not resolve the guest `libva-drm` library’s transitive `libdrm` dependency, so FFmpeg was built without `HAVE_VAAPI_DRM`.

Adding the following build-time linker option to the generated compiler wrapper allowed that check to succeed:

```sh
-Wl,-rpath-link=$root/usr/lib
```

We also added a post-configure assertion so a missing backend fails the build instead of producing an unusable package:

```sh
grep -q '^#define HAVE_VAAPI_DRM 1$' config.h ||
  fail "FFmpeg DRM VA-API backend is required"
```

After rebuilding and installing the package, the decoder tests passed through the normal installed library paths. The `-rpath-link` adjustment only affects build-time dependency resolution; it does not add a runtime search path.

### 2. Build the `mpv` target explicitly

The unrestricted Ninja invocation attempted an auxiliary protocol-list generation step that executed a guest-linked binary in the builder and failed to load `libass`.

Building the target explicitly completed the executable used by the package:

```sh
ninja -C "$build/mpv-build" -j"$jobs" mpv
```

The full project test suite, app assembly, and strict signature validation passed with these adjustments.

## Runtime results

| Area | Observed result |
|---|---|
| Hardware video decoding | HEVC 8-bit, HEVC 10-bit, VP9 8-bit, and AV1 10-bit each produced the expected 120 frames: **480 frames total**. |
| Guest CPU consumption | Hardware decoding used approximately **83–88% less guest CPU time** than software decoding in the short test clips. Software decoding had higher maximum throughput; this does not establish total host CPU or power savings. |
| Installed mpv playback | Confirmed VA-API hardware decoding and playback to completion for 10-bit HEVC, but recorded **7 dropped frames during the four-second clip**. |
| Multisampling | Both the default guest Mesa and the private HDR Mesa exposed four samples and passed actual multisample allocation, rendering/clear, resolve, and readback checks. |
| Native Wayland presentation | Approximately **119–120 fps** from presentation feedback, including a modest rendering workload. This measures the virtual presentation path, not physical panel scanout. |
| HDR | Guest 10-bit output and BT.2020/PQ metadata handoff were observed; direct host presenter tests passed. End-to-end color accuracy remains unverified. |
| Chromium | WebGL 1 and WebGL 2 passed with explicit ANGLE/GLES selection, using VirGL rather than software rendering. |

## Remaining validation

- Audio continuity.
- Firefox.
- A controlled 60 Hz run. One attempted run was restored to 120 Hz by display synchronization, so it has not been counted as 60 Hz coverage.
- Longer playback and soak tests, including dropped-frame measurements.
- Visual HDR correctness.

## Separate build reproducibility finding

Initial guest assembly encountered an Aquamarine/hyprtoolkit ABI mismatch in the available package set. We used a compatible local pair to continue testing. I would track that separately unless it can be tied directly to this change.

Fail-Safe:
Follow-up to my earlier M3 Max testing: we now have a useful SDR-white calibration result and a repeatable HDR-highlight discrepancy to investigate separately.

These observations are from the local integration of this PR, subsequently updated to guest kernel `7.2.5-1-aarch64-ARCH`, with temporary diagnostic instrumentation. They are **not qualification of an unmodified checkout** of the current PR head (`38f7f2b7011002eb72d781147f9eb95960efba62`).

### SDR-white calibration

With Mac display brightness held at **50%**, setting Hyprland's `sdr_max_luminance` to **105** brought ordinary white in the guest into close agreement with ordinary white on the host:

| Comparison | Phone light-meter reading |
| --- | ---: |
| Host ordinary white | 14.2 fc |
| Guest ordinary white, SDR white 105 | 14.2–14.3 fc |

The host/guest difference remained approximately **0.1 fc** when rechecked after HDR video playback. These are relative phone light-meter measurements, not calibrated panel luminance or colorimetry. **105 is a useful calibration for this tested display/setup at 50% brightness, not a proposed universal default or a claim of 105 physical panel nits.** Other brightness levels, display presets, and displays still need separate checks.

### HDR highlights remain a separate issue

Keeping SDR white at 105 preserves visibly brighter HDR output, but the same HDR-white test video in host and guest Vivaldi produced:

| HDR video comparison at 50% host brightness | Reading |
| --- | ---: |
| Guest Vivaldi | 63.7 fc |
| Host Vivaldi | 90.1 fc |

That is approximately **29.3% lower guest output**, or **0.50 stops**, while ordinary white remains matched. An earlier run was similarly around 64 versus 89 fc.

We corrected an initial resolution mismatch before the final readings. The final statistics showed **3840×2160 at approximately 59 fps and PQ/BT.2020** on both sides. A remaining confounder is that the host received **AV1, YouTube variant 701**, while the guest received **VP9, variant 337**. This is therefore not an identical-bitstream comparison, and we have not isolated decoding, guest tone mapping/compositing, or Mac presentation as the cause. Raising SDR white to compensate would discard the successful desktop calibration without resolving that question.

### Digital signal checks

A temporary diagnostic sampled the GL framebuffer immediately before the existing PQ-to-Metal conversion. In controlled, focused white-page captures, requested SDR-white levels produced approximately:

| Requested SDR white | Decoded red PQ signal luminance for near-neutral white samples |
| --- | ---: |
| 140 | 136.4–137.7 |
| 170 | 166.5–168.1 |
| 250 | 244.2–246.5 |

This confirms that changing the setting changes the guest's emitted white signal. Those values are **digital signal luminance, not measured panel nits**. Later neutral bright samples reached roughly 490–512 during the video investigation, but were not tied to a specific frame and should only be treated as a lead. The readback instrumentation can stall rendering, so these diagnostic runs are not performance evidence.

### Persistence and next steps

The calibrated 105 setting is now saved in the local guest's monitor configuration and display-sync override. Live configuration reload and the display-change reapply path retained **10-bit HDR (`XRGB2101010`) and 105**, with no configuration errors. Saving only a one-off `hyprctl` adjustment is insufficient: startup or display synchronization can replace it with the EDID-derived default. This is a local personalization, not an upstream patch; restart validation is still pending.

The useful next investigation is an equivalent, controlled HDR source with known signal values, traced through guest scanout and Mac presentation while retaining the SDR-white control. Current results establish a practical desktop-white match and HDR output above ordinary white, but **do not yet establish host/guest HDR-highlight equivalence or end-to-end color accuracy**.

themartiano:
@nocell @Fail-Safe Very promising indeed!
Works on my M2 Pro.
What do we need to make this production-ready?
From what I've seen we are not querying the underlying hardware to know what formats are supported, that could be good (VP9, AV1, etc).
Also, Codex says Firefox can ditch the decoder 'permanently' after two failures, so we need to check that.
If we manage to get it working on any browser on any Mac, it's the holygrail.

I've pushed some changes to fix build and merged main.

Fail-Safe:
I'll try to circle back around to this sometime soon. Life has been getting in the way of fun as of late. 🙃

PR #172, opened 2026-09-08
[fix] Honor Omarchy terminal preference in xdg-terminal-exec

### LINKED ISSUE #169: Defaults Terminal
I was able to get alacritty and kitty installed, but when I try to make either of them default the settings do not stick.  It keeps defaulting to foot.  I an able to to launch either of them, but can't make them defaults.

Fail-Safe:
## Summary
- Fixes Setup → Defaults → Terminal appearing not to stick after Alacritty or Kitty is installed ([#169](https://github.com/omacom/try-omarchy/issues/169)). Omarchy already writes `~/.config/xdg-terminals.list`; the factory `xdg-terminal-exec` shim always printed and launched Foot, so the Defaults menu (`--print-id`) and Super+Return never followed that list.
- The shim now allowlists Omarchy's supported desktop IDs (Alacritty, Foot, Kitty, Ghostty), skips unavailable terminals, and falls back to Foot. For the factory Alacritty and Kitty wrappers, selection also requires an executable underlying `/usr/bin` package binary. Launching still uses the PATH-resolved wrapper so the VirGL graphics workaround remains active. It does not parse arbitrary user `.desktop` files.

## Existing VMs
Try Omarchy's factory image is a seed: installing a newer app does not rewrite a provisioned disk, so this overlay change does not land on an already-running guest by itself.

Foot still launches there, so the stub can be replaced in place without a factory reset. From a guest terminal, overwrite `/usr/local/bin/xdg-terminal-exec` with `guest/factory-overlay/usr/local/bin/xdg-terminal-exec` from this PR and `chmod 0755` it. Super+Return and Setup → Defaults then follow `xdg-terminals.list`. A factory reset also picks it up on the next first boot.

## Test plan
- [x] Full `make test` on `97f23f2`: 103 guest Python tests, 242 Swift tests, and launcher checks pass. The staged-QEMU lock-inheritance test is skipped because this checkout has no staged QEMU binary.
- [x] Regression coverage uses the factory Alacritty and Kitty wrappers to verify default reporting and fallback with missing/non-executable package binaries, and preservation of wrapper rendering settings when installed.
- [x] Integrated current `main` and resolved the executable-list and guest-verification conflicts.
- [ ] Install Alacritty, set it as default, confirm Setup → Defaults shows Alacritty and Super+Return launches it
- [ ] Repeat with Kitty
- [ ] With only Foot installed, Super+Return still launches Foot
- [ ] On a provisioned VM, replace `/usr/local/bin/xdg-terminal-exec` with the updated shim and confirm the same without factory reset

Fixes #169


themartiano:
Hey, thanks for the PR.

Codex:

missing-terminal fallback is broken. [Line 48](https://github.com/omacom/try-omarchy/pull/172/files) checks whether a command exists, but the Alacritty wrapper remains installed after the package is removed. I reproduced Super+Return exiting with “alacritty is not installed” instead of falling back to Foot. Kitty has the same problem when combined with #173. Check the underlying binary’s availability while retaining the wrapper for launching.

Fail-Safe:
@themartiano Should be in better shape now if you're able to review again. Thanks!

PR #149, opened 2026-09-07
[fix] Force software GL for Alacritty on VirGL

Fail-Safe:
## Summary
- Fixes https://github.com/omacom/try-omarchy/issues/148: Alacritty installs cleanly on aarch64, but the default VirGL/OpenGL path paints an empty transparent window (wallpaper shows through; no text).
- Confirmed locally that `LIBGL_ALWAYS_SOFTWARE=1 alacritty` renders normally.
- Add a `/usr/local/bin/alacritty` PATH wrapper that execs `/usr/bin/alacritty` with `LIBGL_ALWAYS_SOFTWARE=1` when `omarchy.qemu_virgl=1` is on the guest cmdline (same idea as the 1Password ARM installer software-GL shim).

## Test plan
- [x] `python3 guest/tests/verify.py`
- [x] Manual: `LIBGL_ALWAYS_SOFTWARE=1 alacritty` shows text (reporter/dev repro)
- [ ] After image with this fix: Install → Terminal → Alacritty (or `alacritty` from foot) shows a usable terminal without manually exporting `LIBGL_ALWAYS_SOFTWARE`
- [ ] CI green

Fixes #148.

PR #173, opened 2026-09-08
[fix] Force software GL for Kitty on VirGL

Fail-Safe:
Kitty installs successfully but exits before opening a window on the native VirGL guest: EGL fails to create its context, followed by `Failed to create GLFWwindow`. This reproduces when launching the packaged binary directly, independently of terminal preference selection.

Add a `/usr/local/bin/kitty` wrapper following the Alacritty workaround in #149. It sets `LIBGL_ALWAYS_SOFTWARE=1` only when the exact `omarchy.qemu_virgl=1` kernel option is present, then executes `/usr/bin/kitty` with the original arguments. Other guests retain their existing rendering environment. The wrapper reports a missing Kitty package without recursing.

Validation:
- Full `make test` passed with the Xcode toolchain.
- Regression tests cover exact kernel-option matching, missing command-line input, preservation of the non-VirGL environment, argument boundaries, and missing package handling.
- Installed the wrapper in a provisioned disposable ARM64 VM with Kitty 0.48.2-1. Super+Return launched a mapped Kitty window without manually exporting a rendering variable; its process environment contained `LIBGL_ALWAYS_SOFTWARE=1`.
- Captured a rendered Kitty window executing a command and reporting the requested working directory containing spaces.
- Live terminal selection used the shim from #172. This PR is independent and does not include that shim.
- A new factory image was not rebuilt for this wrapper-only change; live validation installed the exact wrapper file into the existing test guest.

Existing VMs can install `guest/native-overlay/usr/local/bin/kitty` as `/usr/local/bin/kitty` with mode 0755. Updating the host app alone does not modify a provisioned disk. Kitty itself still needs to be installed through the normal terminal installer.

Related: #149, #172.


themartiano:
Fixed conflicts, merging with CI failure cuz unrelated

PR #186, opened 2026-09-10
[fix] Move Reset below Launch in the start menu

Fail-Safe:
Reset currently appears above Launch in the start menu, making it easy to hit the destructive action while intending to start the VM. Move Launch above Reset, retaining the 12-point gap between actions and the 8-point gap before the footer.

This changes only the stack order and spacing. Launch remains the Return-key action; the existing reset confirmation, button styling, and enabled states are unchanged.

Validation: `make test` passed, including 242 native tests, 103 guest tests, and shell/storage contracts. The optional virtio-pinch guest ABI test was skipped because `QEMU_PINCH_TEST_BINARY` was not configured. `git diff --check` passed. The change has not been installed in a local app bundle or visually exercised in the running app.

Screenshot:
image


themartiano:
Closing because already included in `v0.4.0`, but nice one

PR #26, opened 2026-08-27
Preinstall Simplified Chinese fonts and Pinyin input

eachann1024:
The factory guest currently ships `noto-fonts` / `noto-fonts-emoji` and no `fcitx5` stack, so Chinese text becomes tofu and there is no Pinyin engine until the user can `omarchy pkg add` extra packages. That extra step is brittle when guest mirrors are broken (see the now-closed #2). Upstream Omarchy already lists `noto-fonts-cjk`, `fcitx5`, `fcitx5-gtk`, and `fcitx5-qt` in `install/omarchy-base.packages`; this restore plus the official Chinese engine matches the [Omarchy keyboard manual](https://omarchy.org/manual/keyboard-mouse-trackpad/).

There is no existing Chinese / locale / fcitx / CJK issue or PR on this repository. This is not an i18n request to `basecamp/omarchy` (English-only product decision).

## What changed

- Request `noto-fonts-cjk`, `fcitx5`, `fcitx5-gtk`, `fcitx5-qt`, `fcitx5-chinese-addons`, and `fcitx5-configtool` in `guest/packages.txt`.
- Update `packages.lock.json` `requestedFileSha256` to match the new list. The resolved package set is **not** refreshed here: that needs the privileged ARM64 builder (`guest/build-container.sh --refresh-package-lock`).
- Generate `zh_CN.UTF-8` in the factory `locale.gen`. Keep `LANG=en_US.UTF-8` so Omarchy menus stay English.
- Seed Pinyin into the default Fcitx5 group for new accounts via `/etc/skel`, leaving `keyboard-us` as `DefaultIM`.
- Document the seed in the root and guest READMEs.
- Extend `guest/tests/verify.py` so the package list, locale contract, and Fcitx5 profile stay pinned.

## Verification

- `python3 tests/test-build-cache.py` — 6 tests passed.
- `guest/test` — 30 unit tests plus the guest contract checks passed, including the new CJK / locale / Pinyin assertions.
- `bash -n guest/scripts/configure-rootfs.sh` and `python3 -m py_compile guest/tests/verify.py` passed.
- Did **not** run a full `make guest` / Docker ARM64 image rebuild (privileged builder + ~20 GB). The next guest rebuild must refresh `packages.lock.json` and confirm the 6144 MiB factory disk still fits CJK fonts.
- Did **not** run `swift test` or `make test`; this change is guest-contract only.

## Existing VMs

[...]

PR #188, opened 2026-09-11
[feat] Restore upstream's CJK font and fcitx5, add Traditional Chinese

### LINKED ISSUE #175: zh-TW support: restore upstream's fcitx5 and Noto CJK, add Traditional Chinese locale and Bopomofo input
🤖 _This issue was written by Claude Opus 5 (an Anthropic AI model) on behalf of @k5953837, who asked it to add Traditional Chinese support and to read the project's contribution rules first. The upstream package list, the package sizes, and the mirror state below were read directly from the pinned Omarchy tree and the live Arch Linux ARM mirrors on 2026-09-09. The branch described under “Status and blocker” passes `guest/test` locally; the factory image itself has **not** been built, for the reason given there._

## Summary

The factory guest cannot display or type Traditional Chinese today. It generates only `en_US.UTF-8`, ships no CJK font, and omits fcitx5, so Chinese text renders as tofu boxes and there is no way to enter it.

The part worth stating up front: **upstream Omarchy already ships the pieces that make this work.** At the pinned release (4.0.2, `346e69e1c`), `install/omarchy-base.packages` contains:

```
35: fcitx5
36: fcitx5-gtk
37: fcitx5-qt
88: noto-fonts-cjk
```

`guest/packages.txt` — the intentionally trimmed ARM64 transaction — drops all four. So this is not a request to add something Omarchy lacks. It is a request to restore what the trim removed, plus the Traditional-Chinese-specific pieces upstream does not cover.

The trim has already cost one bug: #91, where the omitted `fcitx5` binary left upstream's `omarchy-fcitx5.service` in a 203/EXEC restart loop that flooded the journal. #106 fixed that defensively with a `ConditionPathExists=/usr/bin/fcitx5` drop-in — a guard that is self-healing, so installing the package makes it pass with no change to the guard.

## Why this belongs here rather than upstream

Two earlier pull requests proposed Simplified Chinese support (#26, #63). Both were withdrawn by their own authors, and #26's closing comment argued the work "belongs in upstream Omarchy... not the Try Omarchy wrapper."

The package list above is the reason I think that reasoning does not hold. Upstream has already made the call to ship fcitx5 and Noto CJK. The gap exists only inside this fork's ARM64 trim, so it can only be closed here. What upstream does *not* provide, and what is genuinely this project's to decide, is the zh-TW-specific configuration: a Traditional Chinese locale, a font rule that picks TC Han glyph variants over Simplified or Japanese ones, and a Bopomofo engine.

## Proposed change

| Concern | Change |
|---|---|
| Locale | Generate `zh_TW.UTF-8` alongside `en_US.UTF-8` in `configure-rootfs.sh`. **`LANG` stays `en_US.UTF-8` and `KEYMAP` stays `us`** — the locale becomes available, not default. |
| Fonts | Restore upstream's `noto-fonts-cjk`. |
| Glyph variants | New `etc/fonts/conf.d/30-*.conf` preferring `Noto Sans/Serif/Mono CJK TC` when the text language is `zh-tw` or `zh-hant`. Without it, shared codepoints render with Simplified or Japanese forms. |
| Input method | Restore upstream's `fcitx5`, `fcitx5-gtk`, `fcitx5-qt`; add `fcitx5-chewing` for Bopomofo. |
| Input method defaults | Seed `/etc/skel/.config/fcitx5/profile` with `keyboard-us` as the first item and `chewing` as the second, so a user who never presses Ctrl+Space sees no change at all. |
| XWayland | `XMODIFIERS=@im=fcitx` in `environment.d`. `GTK_IM_MODULE`/`QT_IM_MODULE` are deliberately left unset so native Wayland GTK4/Qt6 apps keep using text-input-v3, which Hyprland implements. |
| Chromium | Append `--enable-wayland-ime` to upstream's `/etc/skel/.config/chromium-flags.conf`. Chromium on Wayland ignores input methods without it, and Chromium is the browser this image ships. |

Net effect for an existing English-speaking user: none. The default session language, keyboard layout, and first input method are unchanged.

## Deliberately excluded

- **`fcitx5-chinese-addons`** — hard-depends on `qt6-webengine`, which does not fit the 6144 MiB factory image. #63 hit the same wall. Bopomofo via `fcitx5-chewing` does not need it.
- **`fcitx5-configtool`** — upstream does not ship it either, and it pulls a KDE dependency chain (`kwidgetsaddons` alone is ~15 MiB). Excluded to keep the package set identical to upstream's plus one engine.
- **Raising `image.sizeMiB`** — that value is duplicated as a runtime contract in `macos/run-qemu-gpu.sh` and asserted in `macos/Tests/run-qemu-ssh-contract.test.sh`, so changing it is an architecture discussion, not part of this.
- **Localising the macOS launcher UI** — unrelated to the guest and a separate proposal.
- **Changing the default language** — see above.

## Size

Measured `Installed Size` from `pacman -Si` in an aarch64 builder container:

| Package | Installed Size |
|---|---|
| `noto-fonts-cjk` | 298.79 MiB |
| `fcitx5` | 20.72 MiB |
| `fcitx5-qt` | 2.28 MiB |
| `fcitx5-gtk` | 516 KiB |
| `fcitx5-chewing` | 171 KiB |

Plus new transitive dependencies, of which `libchewing` (6.78 MiB) is the largest; the exact set is pending the lock refresh below.

`guest/scripts/pack-image.sh` requires the staged root to stay under 85% of 6144 MiB. I have not been able to measure the actual headroom yet, for the reason below. If it turns out to be tight, `adobe-source-han-sans-tw-fonts` (38.15 MiB, Traditional Chinese only) is a fallback, at the cost of dropping Japanese and Korean coverage that upstream currently provides.

## Status and blocker

The non-package half of this is implemented and passing `guest/test` on a branch here: locale generation, the fcitx5 profile fragment, the Chromium flag append, the fontconfig rule, the `XMODIFIERS` entry, and ten new assertions in `guest/tests/verify.py` covering the contract — including one that the default `LANG`/`KEYMAP` stay English.

I cannot complete it yet: adding packages requires regenerating `guest/packages.lock.json`, and that is blocked by the two upstream drifts in **#171** (the builder's `rust` pin no longer exists on the ALARM mirror, and no single `aquamarine` currently satisfies both `hyprland` and `hyprtoolkit`). I am waiting on that rather than working around it, since `CONTRIBUTING.md` asks that pinned inputs not be changed independently of their validation code.

## Question

Before I open a pull request: is Traditional Chinese support something you want in Try Omarchy? Given that the packages come straight back from upstream's own list and the default English experience is untouched, I think the surface area is small — but #26 and #63 suggest previous contributors were unsure of your position, and I would rather ask than guess.

If you would prefer a narrower first step — for example restoring only `noto-fonts-cjk` so CJK text renders, and leaving input methods for later — I am happy to split it that way.

k5953837:
Fixes #175.

## Summary

The factory guest cannot display or type Traditional Chinese. It generates only
`en_US.UTF-8`, ships no CJK font, and omits fcitx5, so Chinese renders as tofu
boxes with no way to enter it.

Most of what is missing was never Omarchy's decision. At the pinned release
(4.0.2, `346e69e1c`), `install/omarchy-base.packages` already contains:

```
35: fcitx5
36: fcitx5-gtk
37: fcitx5-qt
88: noto-fonts-cjk
```

`guest/packages.txt` — the intentionally trimmed ARM64 transaction — drops all
four. This restores them, adds `fcitx5-chewing` for Bopomofo, and supplies the
Traditional-Chinese-specific configuration upstream does not: a `zh_TW.UTF-8`
locale, a fontconfig rule that picks TC Han glyph variants over Simplified or
Japanese ones, and a way to choose the language.

That trim has already cost one bug: #91, where the omitted `fcitx5` binary left
upstream's `omarchy-fcitx5.service` in a 203/EXEC restart loop. The
`ConditionPathExists=/usr/bin/fcitx5` drop-in added in #106 is self-healing —
installing the package makes it pass with no change to the guard.

### What changes

| Concern | Change |
| --- | --- |
| Packages | Restore `noto-fonts-cjk`, `fcitx5`, `fcitx5-gtk`, `fcitx5-qt`; add `fcitx5-chewing` |
| Locale | Generate `zh_TW.UTF-8` alongside `en_US.UTF-8` in `configure-rootfs.sh` |
| Glyph variants | New `etc/fonts/conf.d/30-try-omarchy.conf` preferring `Noto Sans/Serif/Mono CJK TC` for `zh-tw` and `zh-hant` |
| Input method | Seed `/etc/skel/.config/fcitx5/profile` with `keyboard-us` first, `chewing` second |
| XWayland | `XMODIFIERS=@im=fcitx` in `environment.d` |
| Chromium | Append `--enable-wayland-ime` to upstream's `/etc/skel/.config/chromium-flags.conf` |
| Choosing it | A **Language** row in the start menu, in the shape the other integrations already use |

### How the choice reaches the guest

The Language row stores a preference the same way Shared folder, Port forwarding
and Resources do, `run-qemu-gpu.sh` turns it into a `tryomarchy.locale=` kernel
argument the way it already derives `tryomarchy.ssh_access=1`, and a guest
oneshot unit writes `/etc/locale.conf` before either login entry point starts.

Leaving the row alone emits no environment variable and no kernel argument, so
an English launch is byte-identical to today. Both sides validate the value
against the locales the image actually generates; the host fails loudly on an
unknown one because that value comes from our own code, while the guest ignores
it silently because the kernel command line does not deserve that trust.

### Beyond #175

#175 proposed the guest-side restore only. The Language row is not in it. I
added it because the guest half alone leaves a user with a working locale and no
way to select it short of editing `/etc/locale.conf` by hand, and because it
needed no new mechanism — it is the same preference → `OMARCHY_QEMU_GPU_*` →
kernel-argument path three existing settings already take. Happy to split it out
if you would rather see the packages land on their own.

### Deliberately excluded

- **`fcitx5-chinese-addons`** — hard-depends on `qt6-webengine`, which does not
  fit the 6144 MiB image. #63 hit the same wall. Bopomofo via `fcitx5-chewing`
  does not need it.
- **`fcitx5-configtool`** — upstream does not ship it either, and it pulls a KDE
  dependency chain for a GUI the seeded profile makes unnecessary.
- **Raising `image.sizeMiB`** — that value is duplicated as a runtime contract in
  `run-qemu-gpu.sh` and asserted in `run-qemu-ssh-contract.test.sh`; changing it
  is an architecture discussion, not part of this.
- **Translating Omarchy's own interface** — the setup wizard and menus stay
  English, and README says so. Upstream has no translation mechanism; those
  strings are hardcoded in its shell scripts.

### On the two earlier attempts

#26 and #63 both proposed Simplified Chinese and were closed by their own
authors. The only maintainer feedback on either was on #26:

> Thanks for pushing this! Before merge, please update the package lock so it
> includes the new packages and builds don't fail.

That PR bumped `requestedFileSha256` without a real resolve. This one's lock is
a genuine refresh: nine resolved entries, no existing version touched.

## Validation

**Package lock** — regenerated with `guest/build-container.sh --refresh-package-lock`
against current `main`, not carried over from an earlier branch. The diff is nine
additions, zero removals, and no change to any existing version:

`enchant` · `fcitx5` · `fcitx5-chewing` · `fcitx5-gtk` · `fcitx5-qt` ·
`libchewing` · `noto-fonts-cjk` · `xcb-imdkit` · `yoga`

Five are requested in `packages.txt`; the other four are transitive
(`libchewing` for the Chewing engine, `xcb-imdkit` for `fcitx5-qt`'s XCB input
method, `enchant` and `yoga` via `fcitx5`).

**Tests** — `./guest/test` passes: 110 Python tests including eight new
`test_locale.py` cases (allowlist parity, symlink refusal, shell-metacharacter
payloads, idempotency), plus the native guest contract with the new zh-TW
assertions. `swift test` passes 254 tests including eleven new pure-model cases
for the preference store and menu presentation.

**Image** — `make guest` completes and the packed root sits at 82.97% of the
6144 MiB budget, inside `pack-image.sh`'s 85% ceiling.

**Booted** — built the image, chose Traditional Chinese in the start menu, and
ran a disposable VM through first-boot provisioning on an Apple Silicon Mac. In
the guest:

```
$ echo "LANG=[$LANG]"; cat /etc/locale.conf; systemctl is-active try-omarchy-locale.service
LANG=[zh_TW.UTF-8]
LANG=zh_TW.UTF-8
active
```

`xdg-user-dirs` renamed the home directories (桌面, 下載, 文件, 音樂, 圖片, 影片,
專案), `eza` printed dates as `11 9月 02:09`, and every glyph rendered — no tofu.
`Ctrl+Space` reached Chewing and `su3` `cl3` committed 你好, after which
`Ctrl+Space` returned to plain US input:

```
$ echo 你好 from omarchy zh-TW
你好 from omarchy zh-TW
```

Leaving the Language row alone emits no kernel argument; `/etc/locale.conf` then
gets the image default, which is also what makes switching back to English take
effect on a VM that has already run in Chinese.

---


k5953837:
CI is sitting at `action_required` — this is my first PR here, so the workflow
needs your approval before it runs. Nothing to fix on my end; flagging it so it
does not look like a stuck check.

For what it is worth, `make test` passes locally on macOS 15 / Apple Silicon:
`./guest/test` is clean (110 Python tests, native guest contract verified), and
`swift test` passes 254 tests. The one exception is
`StartMenuWindowWidthTests.swift:131`, which expects `content.bounds.width == 600`
and gets `601.0` — it fails identically on a clean `main` checkout in this
environment, and CI has it passing on `main`, so it reads as a local font-metric
artifact rather than anything this branch touches.

k5953837:
Rebased onto `main` (`17d7194`) to clear the conflict that #198 introduced.

One hunk actually conflicted: both branches append to `integrationRowViews` in
`StartMenuWindow.swift`. Kept both rows in the order each side established —
`resourceRow, networkingRow, portForwardingRow, immersiveRow, languageRow`.

`run-qemu-gpu.sh` and `VMApplicationController.swift` merged cleanly, and I
re-read both afterwards rather than trusting that: the locale token is still
appended to the same `-append` line, and the launch-configuration chain still
threads every step's environment into the next (`audio → sharing → forwarding →
fullscreen → resources → language → storage`), which is the part a rebase can
silently break without failing to compile.

After the rebase:

- `./guest/test` — clean, 110 Python tests, native guest contract verified
- `swift test` — 273 tests in 61 suites; the four new Language suites pass
- The one failure is still `StartMenuWindowWidthTests.swift:131` (`601.0` vs
  `600`). I re-ran it against a clean `origin/main` checkout in this same
  environment and it fails there identically, so it is local to this machine
  rather than anything this branch does.

CI is still waiting on first-contributor approval.

PR #142, opened 2026-09-05
[fix] Rebuild compatible Hyprland dependencies for factory images

Fail-Safe:
Fresh ARM64 factory builds cannot resolve the rolling repository's mixed aquamarine ABIs: the pinned Hyprland needs `libaquamarine.so=13`, while the current Hyprtoolkit needs `.so=14`.

This rebuilds aquamarine 0.14 and Hyprtoolkit 0.5.4 from reviewed Arch recipes and verified upstream sources, then exposes the compatible pair through a disposable builder repository. Hyprtoolkit uses package release `6.1` to distinguish the rebuild. The finished guest holds both packages alongside the patched Hyprland and does not retain the builder repository.

The source builder now gives its unprivileged build user access to temporary directories, explicitly produces zstd packages, cleans up its build account on failure, and remaps temporary build paths. Library hashes are verified before packages enter the repository. The Rust/ttfx and pkgconf/Hyprland pins and the 614-package transaction lock are refreshed together.

Validation:
- [CI passes on `3f7ec01`](https://github.com/omacom/try-omarchy/actions/runs/34285303029).
- Full `make test` passes with the Xcode toolchain, including new regression tests for the compatible pair, missing archives, and guest/builder package holds.
- Independent source rebuilds reproduce the aquamarine and Hyprtoolkit library hashes; the rebuilt ttfx matches its reviewed hash.
- Full `make build` passes with a new, empty guest-work volume. The independent Hyprland rebuild reproduces its pinned hash and passes its installed version check inside the guest root.
- The app passes strict code-signature validation and the macOS 15 compatibility check.
- An isolated ephemeral VM boots kernel 7.2.3-2, renders first-run setup, and reaches the graphical desktop after setup. The tester completed setup and supplied a screenshot confirming the rendered panel, wallpaper, and welcome notifications. QMP screenshots remain black with the accelerated display, so visual confirmation used the supplied screenshot.

Fixes #171 and extends the factory-build work discussed in #134.

PR #136, opened 2026-09-04
[fix] Hold aquamarine with the pinned Hyprland ABI

Fail-Safe:
## Summary
- Fixes https://github.com/omacom/try-omarchy/issues/134: Omarchy updates fail when pacman tries to install `aquamarine 0.15.0-2` (`libaquamarine.so=14`) while `IgnorePkg` keeps the patched Hyprland that still requires `libaquamarine.so=13`.
- Add `aquamarine` to the existing kernel/Hyprland `IgnorePkg` hold and document the ABI coupling.
- Extend the guest contract test to require the new hold.

## Test plan
- [x] `python3 guest/tests/verify.py`
- [ ] Fresh or reset guest: confirm `/etc/pacman.conf` (via restored `/usr/share/try-omarchy/pacman.conf`) lists `aquamarine` on `IgnorePkg`
- [ ] Run Omarchy update and confirm it no longer fails on the aquamarine / hyprland SONAME conflict

**Note for existing VMs:** this lands in new/reset guest images. Already-provisioned disks still have the old baked `pacman.conf`; they can add `aquamarine` to `IgnorePkg` in both `/usr/share/try-omarchy/pacman.conf` and `/etc/pacman.conf` (or reset) until they pick up a new image.

PR #192, opened 2026-09-11
[fix] Refresh the factory lock for available ARM packages

Fail-Safe:
Factory builds stop before package installation when the rolling ARM repository transaction differs from the checked-in lock. Refresh the lock against current upstream v0.4.0 (`7c7acee`) and current repository metadata.

## Changes

Only seven package versions differ from current `main`:

- `libadwaita`: `1:1.9.3-1` to `1:1.9.4-1`
- `libde265`: `1.1.2-1` to `1.1.3-1`
- `libtirpc`: `1.3.7-1` to `1.3.8-1`
- `linux-aarch64` and `linux-aarch64-headers`: `7.2.5-1` to `7.2.6-1`
- `qt6-declarative`: `6.11.2-1` to `6.11.2-2`
- `tzdata`: `2026c-1` to `2026d-1`

The reviewed Hyprland, Aquamarine, Hyprtoolkit, and Mesa versions remain unchanged. The package list, build specification, and ABI pins are unchanged from current `main`. This changes the factory-image transaction only; it does not update existing VM disks.

## Validation of this revision

- Generated the lock with the project refresh command using fresh isolated repository databases and source rebuilds of the reviewed ABI packages.
- Full `make test` passed: 107 guest tests, 268 native tests, launcher/network/storage checks, and 14 disk-resize tests. The optional virtio-pinch guest ABI test was skipped because no test runtime binary was supplied.

- Built the complete factory guest from a brand-new empty package/source work volume. Package integrity checks, source-built ABI packages, the expected patched Hyprland binary hash, final filesystem checks, image compression, and manifest generation passed.
- This refresh did not rebuild the QEMU runtime, package a new Mac app, or boot the resulting factory image. A fresh-image first-boot check remains release qualification.

Rolling repositories can change again; this refresh restores the transaction available during validation rather than providing a permanent package archive.


jtavin-smp:
Heads up that a fresh refresh no longer resolves, as of today. This lock was generated on 2026-09-15 and CI only checks the lock against `packages.txt` and the spec, never re-running a live resolve, so a green check here does not mean the refresh command still works.

Running `guest/build-container.sh --refresh-package-lock` now fails at the resolve step:

```
:: unable to satisfy dependency 'libaquamarine.so=14-64' required by hyprland
error: failed to prepare transaction (could not satisfy dependencies)
```

Arch Linux ARM `aarch64/extra` today:

| Package | Version | aquamarine soname |
| --- | --- | --- |
| `aquamarine` | 0.15.0-2 | provides `.so=14` |
| `hyprland` | **0.56.2-3** | requires `.so=14` |
| `hyprtoolkit` | 0.5.4-6 | requires `.so=14` |

Upstream is now self-consistent on `.so=14`. The factory is not: `spec.json` pins `abiPackagePins` at `aquamarine 0.14.0-2` (`.so=13`) with `hyprtoolkit 0.5.4-6.1` rebuilt against it, and `supplyChain.omarchy.packages.hyprland` expects `0.56.1-3`, which ALARM no longer carries. `hyprland` is not an ABI pin and is not held during a factory build, so it resolves to 0.56.2-3 and demands the `.so=14` the pinned aquamarine cannot provide.

The 0.14 pin was the right call when ALARM's Hyprland needed `.so=13` (#136, #142). Now that Hyprland has moved, the pin is what blocks resolution rather than what protects it.

I verified this is not specific to any one branch: the same refresh fails identically with an unmodified `main` package list, so it is not caused by a package being added. Practically it blocks any PR that has to relock — #218 is stuck on it.

I have not proposed a fix here because dropping the ABI pins also means re-checking that `patches/hyprland/rounded-border-coverage.patch` still applies to 0.56.2 and that the reviewed Hyprtoolkit rebuild is still wanted, which seems like a maintainer call rather than something to slip into an unrelated PR.

themartiano:
Closing cuz had to updated to build other PRs, we have to solve this eventually. Will take a look once PRs are handled.
Thanks anyway.

PR #195, opened 2026-09-12
fix(guest): add us-acentos keyboard layout option

dyl-joseph:
## Why

Try Omarchy runs on Macs with ANSI keyboards. The onboarding list offered only br-abnt2 for Brazilian Portuguese. That layout puts semicolon on the key labeled slash. ANSI hardware has no AB11 key. This change adds us-acentos. It maps to XKB us with the intl variant. Accents work through dead keys. Slash stays on its printed key.

## Scope

- guest/patches/omarchy/keyboard-us-acentos.patch adds one row to OMARCHY_KEYBOARD_LAYOUTS.
- guest/spec.json registers the keyboard-us-acentos backport.
- guest/tests/verify.py requires the backport and checks patch content.
- guest/packages.lock.json refreshes the ARM64 lock to current repos.

## Tradeoffs

Upstream owns the list. A backport keeps the pinned tree verbatim and applies one reviewed row. A label that names Portuguese and Brazil would read better. The issue requests the English US intl wording. This keeps the requested wording so upstream can adopt the same row.

## Blast Radius

The first boot picker gains one entry. Existing entries keep their order and values. apply_keyboard and input.lua stay unchanged. systemd-firstboot resolves us-acentos through kbd-model-map.

## Verification

- Pinned file has zero us-acentos rows before. It has one after.
- git apply check passes with whitespace error on.
- apply-omarchy-backports.py applies the backport and the digest matches.
- Guest backport unit tests pass with 5 tests.
- guest/tests/verify.py fails on the red commit and passes on green.
- Booted the built guest on an Apple Silicon Mac. Picked English (US, intl. with dead keys) at onboarding. Typed /, ', e for é, ', c for ç, ~, a for ã, ^, a for â, and ', space for a literal quote. All matched.

Fixes #194.

PR #196, opened 2026-09-12
chore(guest): refresh ARM64 package lock

dyl-joseph:
## Why

The guest build fails on main. The package repos moved past the pinned lock. This refreshes the lock to what the repos serve now.

## Scope

- guest/packages.lock.json with 16 version bumps plus new numactl 2.0.19-1. No removals.

## Blast Radius

Guest image contents only. Versions move forward. No config or behavior change is intended.

## Verification

- Generated with guest/build-container.sh --refresh-package-lock.
- Diff matches the drift the guest build reported. Full make guest and make test still need a run.


dyl-joseph:
Superseded. Folded as a commit onto #195 per review.

PR #200, opened 2026-09-13
[fix] Mismatched ICMP replies in macOS NAT networking

Fail-Safe:
Concurrent ping streams in Shared connection (NAT) can receive another request's sequence or payload because libslirp's Darwin ICMP receive path overwrites the saved request before validating the reply. This can produce duplicates, wrong-data warnings, and invalid latency values.

### Before → after

Same standalone workload on the same Mac: three concurrent ICMP echo streams to the same destination, with 180 requests per run.

| Result | Before: original libslirp 4.9.4 | After: patched libslirp 4.9.4 |
|---|---:|---:|
| Requests sent | 180 | 180 |
| Valid replies | 58 / 180 (32.2%) | **180 / 180 (100%)** |
| Mismatched replies | 122 / 180 (67.8%) | **0 / 180 (0%)** |
| Duplicate replies | 0 | 0 |

The mismatched replies are an association failure, not ordinary packet loss: the original path forwarded replies with the wrong sequence or payload. These figures describe the controlled reproduction, not a guarantee of zero loss on arbitrary networks.

Build libslirp 4.9.4 from checksum-pinned source with a Darwin-specific matching patch. A reply must match the pending request before it is forwarded or the socket is consumed. Matching router errors retain their source and fragmentation-needed MTU; explicit extended-ICMP quote lengths bound request matching. Runtime staging now requires the source-built library so it cannot silently replace it with the unpatched bottle. Temporary diagnostic logging is not included.

Validation:
- Fresh-download full QEMU/libslirp runtime build passed, including all five libslirp tests, relocation, signing, and compatibility checks for all 16 runtime Mach-O images targeting macOS 15.
- Complete app assembly passed with a schema-compatible existing factory guest, including deep/strict signature verification and macOS 15 compatibility checks for all 17 Mach-O images. The guest image was not rebuilt for this dependency change.
- Full `make test` passed with Xcode selected, including 242 Swift tests in 54 suites and the portable packet-matching regression tests. Command Line Tools attempts could not load TestingMacros; the Xcode run passed.
- Deterministic Darwin receive test checks that a mismatched reply preserves the pending request, the subsequent matching reply is delivered, and an MTU error preserves guest identity, source, MTU, and checksum.
- Original-library standalone reproduction: 58 valid and 122 mismatched replies out of 180. Patched standalone test: 180/180 valid, no malformed replies or duplicates. Final source-built/staged library: 90/90 larger-payload replies, no malformed replies or duplicates.
- Patched test VM: 180/180 simultaneous replies across 56-, 57-, and 1400-byte payloads; DNS and HTTPS passed. This VM check used the diagnostic candidate before removal of logging; the final staged library was separately exercised as described above.

One earlier large-payload standalone run missed its first three probes; a repeat returned all 90. Their cause was not established. This change addresses reply association, not arbitrary Internet packet loss.

The patch and sanitized reproduction notes are included for review. This PR is independent of bridged networking (#198). No libslirp upstream acceptance or release is implied.


Fail-Safe:
Superseded by #201, which includes all ICMP changes from this PR together with the validated IPv4 UDP reply-translation fix. Both implementation commits remain separate for review, and #201 includes the regression coverage and before-and-after results for each fix. Closing this PR to keep review and merge coordination in one place; the branch is preserved.

PR #218, opened 2026-09-15
Add man-db to the guest image

### LINKED ISSUE #217: Guest image omits man-db, so every omarchy-launch-browser call dumps a Chromium core (upstream omarchy#8897)
## Summary

The guest image does not install `man-db`. `omarchy-launch-browser:12` detects
Mozilla browsers by running the default browser with `--help`, and
Chromium-family browsers implement `--help` by `execlp`-ing `man`. With no `man`
present the exec fails, Chromium calls `PLOG(FATAL)`, and every browser launch
leaves a SIGTRAP coredump. The browser itself then starts normally, so the only
symptom is a crash notification and a core.

This is the upstream bug omacom/omarchy#8897. Filing here because the guest image
is what makes it unconditional: on a stock Omarchy install `man-db` is in
`install/omarchy-base.packages`, so #8897 is an edge case. On Try Omarchy it
fires on 100% of browser launches.

## Evidence

```
$ coredumpctl info 781217
Signal: 5 (TRAP) si_code: TRAP_BRKPT
Command Line: /opt/brave-origin-bin/brave --ozone-platform=wayland ... --help

$ prlimit --core=0 -- /opt/brave-origin-bin/brave --help
[FATAL:chrome/app/chrome_main_delegate.cc:426] execlp failed: No such file or directory (2)
exit 133

$ command -v man            # nothing
$ pacman -Q man-db          # error: package 'man-db' was not found
```

## Why it happens

`guest/packages.txt` (85 entries) installs neither `man-db` nor the full
`install/omarchy-base.packages` set, where `man-db` sits at line 78. Unlike the
fcitx5 case in #91, there is no systemd unit to guard — the probe lives in an
Omarchy shell script, so a `ConditionPathExists` drop-in cannot help.

## Suggested fix

Add `man-db` to `guest/packages.txt`. It is small and neutralizes this until
#8897 lands upstream; `MANPAGER`/`MANROFFOPT` are already set in
`default/bash/envs`, so man pages are evidently expected to work. Closing as
"wait for #8897" is also reasonable — this report is mainly to record that the
guest image is the reason it is not an edge case.

## Related

Same class as #91 (guest omits a package from `omarchy-base.packages` that
enabled machinery assumes). That was fixed with a unit guard; this one cannot be.

## Environment

```
try-omarchy-runtime  4.0.2-1  (guest build 2026-08-31)
omarchy version      4.0.2-1
brave-origin-bin     1:1.95.101-1
kernel               7.2.2-2-aarch64-ARCH, aarch64
host                 macOS on Apple Silicon
```

Filed by Claude Opus 5 (1M context) via Claude Code.

jtavin-smp:
Adds `man-db` to the guest image so Chromium-family browsers can `exec` `man` when probed with `--help`, which Omarchy's browser detection relies on and which currently coredumps.

- Add `man-db` to `guest/packages.txt`

Closes #217.


jtavin-smp:
Superseded by #225, which carries the same change on a branch matching the repository's `fix/` naming convention.

PR #219, opened 2026-09-15
[fix] False kernel reboot prompts on ARM guests

Fail-Safe:
## Summary
Fix the recurring “Linux kernel has been updated. Reboot?” prompt after a no-op `omarchy update` in the ARM guest.

The upstream detector assumes `/usr/lib/modules/<release>/vmlinuz` exists and defaults to an updated kernel if it cannot find a matching image. Arch Linux ARM supplies matching modules without that image at that path, so the prompt repeats even with the direct-boot kernel pinned and already running.

- Recognize package-owned `modules.builtin` metadata alongside the existing `vmlinuz` layout.
- Require a recognized installed release before claiming the kernel changed; any matching running release suppresses the kernel prompt.
- Preserve explicit reboot markers, Hyprland reboot detection, and service/shell restart behavior.
- Deliver through the existing reviewed-backport mechanism, with patch/preimage/postimage SHA-256 pins and guest documentation.

## Validation
- PASS: `python3 guest/tests/test_update_restart_arm_kernel.py` — seven regression tests covering ARM matches and changes, vmlinuz compatibility, multiple kernels, unowned/stale markers, unknown layouts, and other reboot/restart reasons.
- PASS: `python3 guest/tests/verify.py` — guest contract and backport digest checks.
- PASS: fixture SHA-256 independently matches the raw upstream command at pinned commit `346e69e1cec6c4e8924531874af6ba010a1bc99e`; tests apply the actual patch and verify its output digest before executing the complete command with isolated filesystem/process boundaries.
- PASS: `git diff --check`.
- `make test` was attempted on Linux ARM64: 114 guest tests ran, with three failures and three errors in the existing boot-export and native-authentication tests. All six reproduce on unchanged base commit `58cbac5` (boot-export source-size errors and authentication temporary-directory permissions under `/run`). The run stops before the macOS suite.
- Full guest image build was not run because Docker is unavailable in this environment.

## Delivery
This is a factory backport. Existing persistent guests retain their packaged runtime until a separate migration or reset, as documented by the repository; replacing the app alone does not install this fix into an existing guest.


Fail-Safe:
Tightened the detector in 3276e5a to use three explicit states: matched, changed, and unknown. When no recognized package-owned kernel files are found, it now reports “Unable to determine kernel reboot status” on stderr rather than silently treating that case as a match. The diagnostic is independent of the remaining reboot checks, so explicit reboot markers, Hyprland detection, and service/shell restarts continue normally.

Updated the backport checksums and documentation. All seven regression tests pass, including assertions that the diagnostic appears only for unknown layouts and that other reboot reasons remain active. Guest contract checks and git diff --check also pass. Previously documented full-suite and Docker limitations still apply.

themartiano:
man nice, this is so welcome

PR #228, opened 2026-09-18
Back off between QMP startup validation attempts

NimbleAINinja:
## What this changes

`QMPVMHostSleepController.init` validated QEMU's monitor with three connection attempts issued back to back with **no wait between them**. The existing comment already acknowledged that "the launcher sees the socket before QEMU finishes initializing its monitor" — but three attempts inside the same instant do not give it any time to finish. This spaces them: five attempts at 0, 50, 100, 200 and 400 ms, bounded at 750 ms, spent only on the failure path.

A monitor that never answers still fails startup with the same error, so a broken monitor is not mistaken for a slow one.

## Honest scope

This was written while investigating a startup failure:

> The virtual machine started, but Try Omarchy could not enable safe Mac sleep. Please close and reopen the app. (I/O failure: cannot connect to QMP socket: Connection refused)

**It did not fix that failure**, which recurred with this change installed. The symptom appears only on the first launch after installing a new build and has not reproduced since (four subsequent launches, via both the terminal and LaunchServices, all succeeded), so the underlying cause is still unidentified — a persistent `ECONNREFUSED` suggests nothing was listening on the socket at all, rather than a monitor that was merely slow to accept.

I am offering this only as a correctness improvement to an obviously unpaced retry loop, not as a fix for that bug. Happy to close it if you would rather wait until the real cause is known.

Worth noting for whoever picks that up: when validation fails, `VMApplicationController.failHostSleepControlSetup` sends QEMU `SIGTERM`, so a false negative here destroys a running VM and leaves an empty guest console log.

The wait blocks the main thread. That is confined to the retry path, but it is why the total is kept under a second; happy to move it off the main thread instead.

## Validation

- `make test` passes.
- Updated the retry-limit test to assert the attempt count and the delay schedule.
- Added a test covering a monitor that refuses twice before accepting.


NimbleAINinja:
Closing in favour of #229, which fixes the actual cause. The failure was the launcher printing Ready before QEMU's main loop was accepting on its backlog-1 QMP socket; once a probe times out there, the queue stays full and every retry is refused regardless of pacing, so this backoff cannot help. Details and reproduction in #229.

PR #236, opened 2026-09-19
Experimental host USB passthrough, plus two build fixes

vcanuel:
## What this is

An experimental opt-in that hands **one** Mac USB device to the guest -- a new
row on the start menu, a picker sheet, and the launcher and runtime changes
behind them -- plus two independent build fixes found while getting the runtime
to compile.

It is offered as a proposal rather than a finished feature, but not because the
mechanism is unproven. Passthrough works end to end on macOS, including USB mass
storage, and the measurements are below. What it needs is a decision this branch
cannot make: **libusb will only take a device away from macOS for a process that
holds `com.apple.vm.device-access` or runs as root**, and choosing between those
two belongs to whoever ships the app.

## The start menu

A **USB device (experimental)** row joins the integration rows, between port
forwarding and immersive mode, with the same shape as the shared folder row:

- **Choose…** opens a sheet listing every non-hub device attached to this Mac
  through IOKit, as `Name · vid:pid` so it matches what `lsusb` prints in the
  guest. A saved device that is currently unplugged stays selectable as
  *Not connected: …*, so reopening the sheet cannot silently forget it.
  Saving turns passthrough on; the sheet says the choice applies on the next
  start, that macOS keeps drives, keyboards, audio, video and iPhones for
  itself, and that replugging while Omarchy runs can freeze the window for up
  to 30 seconds.
- **Turn On** / **Turn Off** keeps the chosen device without reselecting.
- The detail line states the saved device and whether it is plugged in right
  now (*"… is not plugged in. Omarchy will start without it."*). With a
  connected device on, it says the device appears in Omarchy but macOS keeps
  its data -- never that Omarchy takes it over, and a test asserts that.
- When `OMARCHY_QEMU_GPU_USB_HOST` is set in the environment the row shows
  the value, reports that the environment owns the choice, and disables the
  buttons that would not change it -- the treatment the VM Location row
  already gives `OMARCHY_QEMU_GPU_STATE_ROOT`.

Two whole-window layout tests render the row inside a real `StartMenuWindow`,
with and without a device, and assert its text stays clear of the button column.

## The feature

The launcher adds one xHCI controller and one `usb-host` device bound to the
chosen vendor and product identifiers, which survive the re-enumeration a
device performs when it is unplugged, or a phone when it is unlocked. A bus
address does not, which is why it is not used.

One device rather than every device: `usb-host` claims a device away from macOS
for as long as the VM runs, so a blanket grant would hand the guest this Mac's
dock, keyboard and display controls the moment it starts. Hubs are filtered out
of the picker for the same reason.

`OMARCHY_QEMU_GPU_USB_HOST` overrides the saved choice with a raw `usb-host`
property list for scripted launches (`vendorid=0x05ac`, or
`hostbus=1,hostaddr=6`). The value is validated against a strict pattern rather
than pasted onto the command line, and the row reports when the environment owns
the choice and disables the buttons that would not change it — the same
treatment `OMARCHY_QEMU_GPU_STATE_ROOT` already gets on the VM Location row.

The runtime needed libusb, which `--without-default-features` left out.
`libusb` 1.0.30 joins the pinned `arm64_sequoia` bottle set with its checksum,
`--enable-libusb` joins configure, and `lib/libusb-1.0.0.dylib` joins the
runtime manifest so the existing relocation, signing and Mach-O validation cover
it unchanged.

## What was measured

On an Apple Silicon Mac, with the runtime rebuilt from this branch:

- configure reports `libusb : YES 1.0.30` → `enabled`, and builds
  `hw_usb_host-libusb.c.o`
- the staged QEMU exposes `usb-host` and `qemu-xhci`; `libusb-1.0.0.dylib`
  relocates to `@executable_path/../lib/`, signs, and passes the 17-image
  Mach-O validation
- with no opt-in the command line gains zero USB arguments
- with an opt-in it gains exactly `-device qemu-xhci,id=omarchy-usb` and
  `-device usb-host,bus=omarchy-usb.0,id=omarchy-usb-host,<properties>`
- a value carrying extra QEMU options is refused

## What privilege decides

libusb's Darwin backend takes exactly two kinds of caller and turns every other
one away before the kernel is ever asked:

```c
/* libusb/os/darwin_usb.c, darwin_detach_kernel_driver() */
if (darwin_has_capture_entitlements ()) {   /* com.apple.vm.device-access */
  ...
} else {
  if (0 != geteuid()) {
    usbi_warn (ctx, "USB device capture requires either an entitlement "
                    "(com.apple.vm.device-access) or root privilege");
    return LIBUSB_ERROR_ACCESS;
  }
}
```

**Unprivileged, which is how the app ships**, a device macOS drives enumerates
in the guest but never gets a configuration: `lsusb` lists it while
`bConfigurationValue` stays empty, and no `sda`, input device or camera appears.
A standalone libusb program linked against the shipped dylib reproduces it in
isolation — `libusb_detach_kernel_driver` and `libusb_claim_interface` both
return `LIBUSB_ERROR_ACCESS` in 0.00 s. Surveying all sixteen USB devices
attached to the test Mac found an Apple driver bound to every one, docks
included through `AppleUSBHostBillboardDevice`, so the set that works today is
narrow: hardware macOS has no driver for at all, such as SDR receivers, JTAG
probes and vendor-specific boards.

**As root, this same build takes a driver-claimed device completely.** With a
SanDisk 3.2Gen1 stick, a class macOS drives through `IOUSBMassStorageDriver`:

```
# inside the guest
$ lsblk
sda      57.3G disk
|-sda1     200M part vfat  EFI
`-sda2      57G part vfat  VCAUSB  /run/media/vcanuel/VCAUSB
$ basename $(readlink /sys/bus/usb/devices/2-1:1.0/driver)
uas
$ cat /sys/bus/usb/devices/2-1/speed
5000
$ dd if=/dev/urandom of=/run/media/vcanuel/VCAUSB/t.bin bs=1M count=256 conv=fsync
268435456 bytes (268 MB, 256 MiB) copied, 4.17614 s, 64.3 MB/s
```

The guest binds `uas` at SuperSpeed, the volume automounts on the Omarchy
desktop, and writes land on the stick. **No code in this branch changes between
the two cases** — only the effective uid of the process calling libusb.

Neither door is free, which is why this is a maintainer decision and not a
commit here. `com.apple.vm.device-access` is restricted: Apple grants it per
developer team, delivers it in a provisioning profile, and DTS describes it as
intended for virtualization apps shipping on the Mac App Store. Root is not a
free alternative either, because libusb reads the effective uid of the *calling
process* — there is no split like `qemu-bridge-helper`, so it would be QEMU
itself running as root with a GUI attached. libusb/libusb#1014 tracks the same
question upstream. The README's *Passing a USB device to Omarchy* section
carries all of this -- the libusb excerpt, both measured cases and what each
door costs -- so a user who wonders why a drive does not show up finds the
answer there rather than in a closed PR.

## Two failure modes worth knowing

**The main loop blocks on capture.** Every attempt goes through libusb's
`darwin_reenumerate_device()`, which polls an atomic for up to
`DARWIN_REENUMERATE_TIMEOUT_US` — ten seconds — on whichever thread called it,
and QEMU calls it from the main loop. A QEMU-side detach blocked it for 10.08 s,
and one physical replug stalled the guest's USB enumeration for 29.8 s (guest
kernel timestamps 34.98 → 64.77). The guest keeps computing; the window stops
redrawing and stops accepting input until the timeout expires. A successful
capture costs about 5 s of the same at startup.

**An unclean exit wedges the device.** After QEMU was killed mid-reset, every
bulk transfer to the stick timed out — from QEMU and from a plain libusb program
alike, both as root, with `libusb_claim_interface` still reporting success. One
physical unplug and replug restored it, and the same program then completed a
bulk-only SCSI INQUIRY in 20 ms. This cost a long detour here: the same symptom
was first read as a UAS streams problem, and `usb-storage.quirks=…:u` did switch
the guest to bulk-only without fixing anything. Both are in the README so the
next person does not repeat it.

## Two independent build fixes

Both stand alone and are useful without the feature.

**`make runtime` could not build.** QEMU's `mkvenv` runs offline, and the pinned
`wheel` 0.48.0 resolves `packaging` at install time, which no macOS system
Python ships:

```
ERROR: No matching distribution found for packaging>=24.0
*** Ouch! ***
```

`packaging` 26.3 is now vendored beside the other wheels, pinned by SHA-256 and
verified the same way.

**`make guest` hid why it failed.** `resolve-package-lock.py` reported only
stderr, so a failed resolution surfaced as a bare `error: failed to prepare
transaction (could not satisfy dependencies)` while the line naming the package
went to stdout and was dropped. Both streams are reported now, which is how the
real cause became visible:

```
:: unable to satisfy dependency 'libaquamarine.so=14-64' required by hyprland
```

That drift is real and blocks `make guest` right now: `extra` moved aquamarine
0.14 → 0.15.1-1 (soname 13 → 14) and hyprland 0.56.1-3 → 0.56.2-3, so the
reviewed `aquamarine` ABI pin shadows the copy the newer hyprland requires. It is
**not addressed here** — it is a supply-chain review that belongs in its own
change. Worth noting that `extra` is self-consistent again (aquamarine 0.15.1 +
hyprtoolkit 0.5.4-6 + hyprland 0.56.2-3 + guiutils 0.2.2-3 all agree on soname
14), and that `patches/hyprland/rounded-border-coverage.patch` applies cleanly to
the 0.56.2 sources.

## Testing

`make test` passes: 281 Swift tests in 58 suites, plus every shell and Python
contract. New coverage is 11 tests for the preference, launch environment,
environment override and presentation rules, 2 for the row's rendering inside a
real `StartMenuWindow`, and a `qemu-usb-passthrough.test.sh` contract asserting
the default command line stays free of USB devices.

`macos/build-qemu-gpu-runtime.sh` was run end to end; the pinned libusb bottle
checksum was taken from the Homebrew `arm64_sequoia` manifest and the archive was
downloaded and verified before pinning, as was the `packaging` wheel from PyPI.

The root run used a throwaway APFS clone of the app bundle and of the VM state
root, so nothing outside `/private/tmp` was written. Two things had to be worked
around and are worth knowing if you reproduce it: the storage layer correctly
refuses a workspace owned by another uid, and `zstd` running as root copies the
source file's ownership onto the disk image it expands, which lands a user-owned
file in a root-owned workspace.

Two AppKit notes: the start-menu rendering tests had to join the existing
serialized suite, because two `StartMenuWindow`s laid out concurrently on the
main actor disturb each other's reported content size; and the USB row asserts
its own two detail lines stay clear of the button column rather than re-asserting
the menu-wide width the storage test already owns.

## Not addressed

- the Arch Linux ARM ABI drift above
- choosing between `com.apple.vm.device-access` and a root mode, and any code to
  support either; both are maintainer calls
- the two libusb-side problems above, which look like upstream bugs rather than
  anything to fix in this repository
- a raw `/dev/rdiskN` attached as virtio-blk, which would reach a USB drive
  without any of this, but is a different mechanism and a different discussion


themartiano:
This would be a great addition. Makes a big difference for daily usage.
What is missing so we can get that production-ready? Anything I can help with?

Codex reported these issues:
- **Common devices don’t work in the shipped privilege model.** Capturing devices already controlled by macOS requires the restricted entitlement or root. Neither supported path is implemented. The privilege restriction is confirmed in [libusb’s source](https://github.com/libusb/libusb/blob/v1.0.30/libusb/os/darwin_usb.c#L2664-L2684).
- **Lifecycle reliability remains unresolved.** The author reports replugging can freeze the window for roughly 30 seconds, and an unclean exit can leave the device unusable until physically reconnected. These are documented, not fixed. [PR evidence](https://github.com/omacom/try-omarchy/pull/236)
- **Selection doesn’t identify a physical device.** The picker collapses devices sharing vendor/product IDs into one entry; QEMU can take whichever matching unit it finds first. Two identical drives or development boards cannot be distinguished. [Selection code](https://github.com/omacom/try-omarchy/blob/8d25fc37a29d49de35163721ed026b94c7e60c5c/macos/Sources/OmarchyVMHelper/USBPassthrough.swift#L115-L117)

vcanuel:
Thanks! Pushed a fix for the selection issue: identical devices are now listed separately and pinned by `hostport` (+`hostbus`) when a twin is attached; if QEMU can't tell them apart (bus 0 is a wildcard), the launch passes neither. Also merged `main` to clear the conflicts.

The two remaining blockers aren't fixable in this PR alone:
1. **Privilege** — the real unlock is `com.apple.vm.device-access`, which Apple grants per team via a provisioning profile. That's the biggest help you could give: requesting it for the project's Team ID. Running QEMU as root is the only alternative and I don't think we should ship that.
2. **Lifecycle** — the freeze comes from libusb's `darwin_reenumerate_device()` blocking QEMU's main loop. Fixing it means patching QEMU to do the capture off the main loop; I'd rather do that as a follow-up once the entitlement path is settled, since it changes what the capture path looks like.

themartiano:
I'll handle `com.apple.vm.device-access`, see if we can get approval.
Is disconnect / reconnect working smoothly? 
We also need smooth UX for connecting a new device etc. If UI says A-OK it must be working inside the VM.
With `com.apple.vm.device-access` I think we should be able to allow users to connect any devices from any ports.
Also I've fixed two issues I found and pushed.
Bit of a dump, LMK @vcanuel 
Would love to include this in the next release if possible

vcanuel:
Thanks for picking up `com.apple.vm.device-access`, and for the two fixes! I tested the branch on an Apple Silicon Mac with a SanDisk 3.2Gen1 stick, first unprivileged (as shipped), then with QEMU running as root to stand in for the entitlement.

**Unprivileged (as shipped)**

| Scenario | Result |
|---|---|
| Stick on a direct Mac port (bus 2) | ✅ in guest `lsusb`; macOS keeps the data, as documented |
| Stick behind a hub / dock (bus 0) | ✅ after 0489561; refused before |
| Unplug / replug while the VM runs | ✅ no freeze, device leaves and comes back |
| Move to another port | ✅ start menu flags the missing port, choice is kept |

**As root (stand-in for the entitlement)**

| Scenario | Result |
|---|---|
| Capture of a macOS-driven drive | ✅ mounts on the Omarchy desktop at boot |
| Write speed (`dd` 256 MiB, `fsync`) | ✅ 72.7 MB/s, `usb-storage` |
| Unmount, unplug, replug while the VM runs | ✅ back and remounted in ~5 s, window responsive |
| Same with xHCI streams on (`uas`) | ❌ first command after each connection times out after 30 s; mounts only after two resets, ~60 s |

**Two fixes pushed**

- **0489561 — `hostbus=0` is now exact.** With "Keep USB selections bound to the chosen port", every bus-0 device was refused, because upstream QEMU reads `hostbus=0` as "any bus". But libusb's Darwin backend numbers the first controller 0, and on my Mac that's the controller behind *every* dock and hub port, so almost nothing was selectable, while the picker still offered it. A small QEMU patch (`qemu-usb-host-exact-bus.patch`) moves "any bus" to `UINT32_MAX`; the launcher refuses a runtime without it, and the app pins bus 0 like any other bus.
- **442247e — `qemu-xhci,streams=off`.** With streams on, the guest picks `uas`, and bulk streams don't seem to make it through `usb-host` on macOS. With them off, Linux falls back to `usb-storage` by itself (no guest config) and the stick is ready ~1.2 s after connection. I A/B-tested both on the same stick and port.

**Notes**

- **The freeze did not reproduce**, even as root with a real capture. The README now says "can freeze" and that when the libusb re-enumeration timeout actually triggers isn't pinned down yet. Worth re-checking once the entitlement build exists.
- **Don't use Finder's Eject before handing a drive over.** It sends the stick a SCSI stop, and the stick then reports no medium in the guest (`[sda] Media removed`) until physically replugged. `diskutil unmountDisk` works. The picker could say so, or the app could unmount the volume itself before capture.
- The `libusb_kernel_driver_active: -5 [NOT_FOUND]` lines on stderr are harmless: QEMU probes all 16 possible interfaces and the stick has one. They can be silenced with a one-line QEMU change if you want.

**Unrelated heads-up:** a fresh guest build currently fails on `main` too. Arch Linux ARM mirrors rolled 5 packages past `packages.lock.json` (bubblewrap 0.13.0-1, expat 2.8.5-1, ffmpeg 2:9.0.2-1, gzip 1.15-1, pcsclite 2.5.2-1); `guest/build-container.sh --refresh-package-lock` produces exactly those bumps. I kept it out of this PR.

